PR #256, opened 2014-12-28
Fixes for " Sleep and join" example to compile

Chris00:


mor1:
(personally i'd vote for just removing `mirage run` altogether. it's never
worked properly, and in practice, without better control over sudo
environment and automatic management of the `.xl` file, it seems never
likely to work...)

On 31 December 2014 at 19:36, Anil Madhavapeddy <notifications@github.com>
wrote:

> Thanks! We need to look into the role of mirage run, since it's never
> quite worked (it's easier to run the binaries or unikernels directly). I'll
> review and merge tomorrow when I'm back online properly...
>
> --
> <https://github.com/mirage/mirage-www/pull/256#issuecomment-68464375>.
>



-- 
Richard Mortier
mort@cantab.net

avsm:
Many thanks for the patch!  An issue on github.com/mirage/mirage/issues regarding removing/fixing `mirage run` would be appreciated...

Chris00:
Done so.

PR #264, opened 2015-01-07
Some fixes to libvirt-on-cubieboard, mainly formatting

infidel:
I eventually got through most of this procedure after plenty of sweating: 20% due to libvirt TLSSetup, 20% due to my board failing to boot after the rc.d additions, and the rest due to the hot weather here.

I still haven't got libvirt to start at boot time yet but it's a low priority for me, as the only reason I installed libvirt was to try out jitsu. It might be nice to have libvirt built into the SD card image if there is a way to automate the TLS setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 /src/mir-main
 /src/main.native
 /src/www.xl
+/src/www.xe
+/src/www_libvirt.xml
 
 /src/Makefile
 /src/main.ml
@@ -21,3 +23,5 @@
 /src/fat*.img
 /src/make-fat1-image.sh
 /src/make-fat2-image.sh
+
+.*.swp
--- a/tmpl/wiki/libvirt-on-cubieboard.md
+++ b/tmpl/wiki/libvirt-on-cubieboard.md
@@ -1,91 +1,131 @@
+#### Bringing up the board
+
 1. Download and unpack the image for your board.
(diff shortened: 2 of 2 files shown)


avsm:
I wonder if it's easier just to port jitsu to use libxl directly...

infidel:
Yeah, if it's intended to run on the same host then that would be simpler.

I forgot to mention that I was also unable to build mirage-skeleton/xen/static_website+ip, so I tested with static_website (with a static IP manually added to config.ml), but I'm guessing you knew about that. I can submit an issue if not.

MagnusS:
When we update xen-arm-builder to use ubuntu 14.10 we can use the libvirt package that comes with ubuntu - libvirt in 14.04 doesn't have xen support enabled. Until then there is an install script on the SD card in /root/scripts/install_libvirt.sh that downloads and installs libvirt with Xen support (based on these instructions). TLS is not needed as Jitsu can connect to libvirt over the unix socket when it runs on the same host. We are also adding libxl support in the xenlight branch of Jitsu.

infidel:
OK thanks, doing it with the script and without TLS would have been easier. Does it imply that this page should be changed to document it that way, or are the manual installation instructions still useful for some people?

avsm:
I think the page should probably be changed to document that script -- it's easier just to link to it than let this one get out of date.  Current pull req is an improvement so I merged it, thanks!

PR #266, opened 2015-01-15
Add notes from call on 2015-01-14

amirmc:
The parts on TLS should be reviewed

### DIFF
--- a/tmpl/wiki/weekly/2015-01-14.md
+++ b/tmpl/wiki/weekly/2015-01-14.md
@@ -1,0 +1,149 @@
+14th Jan 2014: Project name, 4.02.1 support, TLS on Xen and Error handling
+
+### Agenda ###
+
+- "Mirage OS" or "MirageOS"?
+- Pull OCaml runtime/bindings into packages / OCaml 4.02.1 support
+- TLS on Xen
+- Error handling in Mirage
+- Planning for 3.0
+
+
+Attendees: Amir Chaudhry, Thomas Gazagnaire, David Kaloper, Thomas Leonard,
+Jon Ludlam, Anil Madhavapeddy, Hannes Mehnert, Richard Mortier, Dave Scott
+and Magnus Skjegstad
+
+
+### Notes ###
+
+#### "Mirage OS" or "MirageOS"? ####
+
+We need to settle on how we refer to the project and then be consistent in its
+use. In the early days it was referred to variously as 'Mirage', 'Open
+Mirage', 'Mirage OS' and 'MirageOS'.  When we became a
+[Xen Incubated project][incubation], we had a brief discussion about the
+'official' name (for press releases etc) and we settled on 'Mirage OS'. 
+However, we haven't always been consistent with the way we use the term.  So
+far all the official materials such as new releases, have referred to
+"Mirage OS" (with a space), whereas on Twitter and our slides we see 'MirageOS'
+(without a space) being used. We should stick to only one of these.
+
+**Why does this matter?** - As we attract more users and raise more awareness,
+being consistent in how we represent ourselves will begin to matter more. The
+name is just one of these aspects and even though it may seem trivial, it *is*
+the name of the project we're discussing here. At this stage, there should be
+no confusion about how it's written.
+
+After a brief discussion, it's clear that most of the team are less concerned
+about the exact choice rather than being consistent. Only Amir and Anil had
+opinions to share and they both proposed that we use 'MirageOS' from now on.
+The reasons are that users have independently used this form in tweets and
+posts, it is easier to search for, and it's better to go with the flow as we
+still can.
+
+Overall, we unanimously decided on two things: (1) to move forward with
+'**MirageOS**' as the official name of the project and (2) be much more
+consistent in using it on our written material.
+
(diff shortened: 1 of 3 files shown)


tmpl/wiki/weekly/2015-01-14.md
```diff
@@ -0,0 +1,149 @@
...
+[incubation]: http://lists.xenproject.org/archives/html/mirageos-devel/2013-01/msg00081.html
+
+#### OCaml 4.02.1 support ####
+
+Anil's working on this ([mirage/mirage-platform#115][115]) and thanks ThomasL
+for his patches. Hasn't had time to merge yet but will do soon. Might not
+rebase it and hopes that won't be too much of a problem for anyone.  In
+general we shouldn't do that but this one would take a while to rebase and
+would prefer to merge quickly.  The functor inline patch from 4.02.1 will have
+big performance boost but we should clearly state that we're not dropping
```
amirmc: I may have misheard the comment about the functors patch and perf boost

tmpl/wiki/weekly/2015-01-14.md
```diff
@@ -0,0 +1,149 @@
...
+ThomasL started an [email thread][tls-thread] about getting the TLS stack
+working on Xen.  He's submitted a number of patches for review to different
+libraries and the changes needed are:
+
+1. Add generic error handling for FLOWs, so we can propagate errors reliably.
+2. Fix the page alignment requirements for Netif.
+3. Add TLS support to conduit.
+
+TLS now works if you apply lots of patches and works pretty well (half the
+speed of normal TLS?). Issue now is how to merge without breaking things as
```
amirmc: I didn't quite catch this either. @talex5, care to correct me?

tmpl/wiki/weekly/2015-01-14.md
samoht: the inline patch is not in 4.02.1 yet but it applies cleanly on trunk. The work done to port Mirage to 4.02.1 makes the port to trunk and to the inline branch of trunk trivial (hopefully).

tmpl/wiki/weekly/2015-01-14.md
```diff
@@ -0,0 +1,149 @@
...
+website and written materials. Anil suggested that we may start using a
+different domain for the site too ([mirage.io](http://mirage.io)).
+
+[incubation]: http://lists.xenproject.org/archives/html/mirageos-devel/2013-01/msg00081.html
+
+#### OCaml 4.02.1 support ####
+
+Anil's working on this ([mirage/mirage-platform#115][115]) and thanks ThomasL
+for his patches. Hasn't had time to merge yet but will do soon. Might not
+rebase it and hopes that won't be too much of a problem for anyone.  In
```
talex5: If you want, send it to me at the end and I'll have a go at rebasing it. I do find the history useful from time to time, when trying to understand why something is the way it is.

tmpl/wiki/weekly/2015-01-14.md
talex5: It's half the speed of not using TLS, which is good news!

tmpl/wiki/weekly/2015-01-14.md
```diff
@@ -0,0 +1,149 @@
...
+[xentropyd]: https://github.com/mirage/xentropyd
+
+
+#### Error handling in Mirage ####
+
+ThomasL has been working on this. To make conduit `FLOW` wrap the TLS flows we
+need a way to propagate the errors sensibly, so has added a type to `FLOW`
+that can do this. It might be nice to have an algebraic data type to describe
+an error type, or if we wait a few months until we switch to 4.02.1, then we
+can just use [Open Types][opentypes].  The nice thing with current patch is
```
talex5: We already have algebraic data types for errors, which is good if you want to match on a particular error from a particular concrete implementation (e.g. `FAT_index_corrupted` or something). In theory, this is slighly more correct than using exceptions because error returns don't propagate automatically. e.g. if you get `Refused` from TCP then you know it was the TCP connection itself that was refused, not some other refusal from another function TCP called that happened to throw the same exception. In practice, this doesn't seem to be a problem except when using very generic exceptions like `Not_found`.

I guess using open types means we can list some common errors in the interfaces but have implementations provide more, which may be useful but doesn't solve my problem: when you *don't* want to handle specific exceptions we need a way to get a human-readable string for display (and this is the common case).

So:
* Replacing error codes with exceptions (which I proposed originally) would work most of the time and would simplify the code greatly, although it's slightly less strict than the current system. On the other hand, the current system encourages people to throw away errors (look at any `main.ml` for examples!) so I'm sure it would be an overall improvement. It would require API changes everywhere, however.
* Adding an `error_message` function (which my current patches do) means that people can still handle errors as before when needed, but can also always get a human-readable string too, so there's never any excuse for throwing away diagnostic information. You do still need to write error handling code everywhere when you just want to propagate exceptions, though, which clutters the code, and old code will continue to discard useful information for some time to come.

avsm:
@amirmc: do you want to push @talex5's comments and I can merge?

amirmc:
Been distracted with other writing.  Will do it later tonight and comment when ready.

amirmc:
Ready to merge from my point of view.

Issue #156, opened 2014-07-08
Add Planet feed to website


amirmc:
A blog roll of contributors should be added to the site.

PR #272, opened 2015-01-26
Static News section

voila:
This replaces the existing blogroll with a news section similar to ocaml.org
(See https://github.com/mirage/mirage-www/issues/156)

This retrieves the RSS feeds and create a tmpl/news.html which lists entries chronologically
(it only supports RSS 2 and Atom, right now)
This runs when mirage-www is built.

In the future, it would be nice to have a summary for each entry.

I am new to this, comments and suggestions are welcome !

screen shot 2015-01-27 at 10 47 26 am


avsm:
awesome! Moving a variant of this onto the front page would be really nice later (along with last 10 releases as well, just like opam.ocaml.org)

voila:
It's done, here: https://github.com/voila/mirage-www/commit/1bfbf8b2191210547a4c6c31c6a6599aae946723

screen shot 2015-01-28 at 10 25 13 am

voila:
The releases changelog is done by hand. I guess the 'Latest Releases' could also be a hand-written Markdown file (albeit trimmed-down) ?

samoht:
Looks good to me, thanks!

voila:
I've merged this into my master branch and I will do a new pull request from there

PR #274, opened 2015-01-29
[RFC] Add page explaining error handling in Mirage

talex5:
Although written as if it's a final design, it's intended only as a starting point for discussion, to find out what we do and don't agree on. Please add comments, information about successful approaches you've seen, etc.

### DIFF
--- a/tmpl/wiki/error-handling.md
+++ b/tmpl/wiki/error-handling.md
@@ -1,0 +1,384 @@
+[ This DRAFT document is intended to explain how error handling should work in Mirage.
+  Currently, it's just my (Thomas Leonard's) opinions.
+  Please add comments on the bits you disagree with.
+]
+
+#### Goals
+
+POSIX and C generally make for very poor error handling:
+the information reported to the user is just a small number stored in `errno` and
+errors are easily ignored accidentally.
+
+With Mirage, we have the opportunity to provide something much better.
+OCaml's structured types and exceptions can provide rich diagnostic information that is readable to humans and to machines,
+and its exhaustiveness checks mean we can force callers to consider errors where necessary.
+
+There are three main aspects to consider:
+
+- Providing diagnostic information to humans.
+- Allowing programs to detect and handle certain errors specially.
+- Indicating failure in order to abort or roll back the current operation.
+
+Note that the last two points are very different.
+Exceptions can be thrown at any point in an OCaml program and robust code must be prepared to handle this.
+In particular, this (fictional) code is *wrong*:
+
+```
+  let gntref = Gntshr.get () in
+  match_lwt f gntref with
+  | `Ok () -> Gntshr.put gntref; return (`Ok ())
+  | `Error _ as e -> Gntshr.put gntref; return e
+```
+
+If `f` throws an exception, the grant ref will be leaked.
+The correct pattern is:
+
+```
+  Gntshr.with_ref f
+```
+
+This will release the resource whether `f` returns success, returns an error code or raises an exception.
+
+#### The old/current system
+
+There are three methods currently used in (Lwt) Mirage code to signal errors.
+First, the code can raise an exception:
+
+```
+let load x = if ...
+  then raise (Permission_denied "Cannot read ...")
+  else return ...
+```
+
+Secondly, it can evaluate to a failed thread on error and a returned one on success:
+
+```
+let load x = if ...
+  then fail (Permission_denied "Cannot read ...")
+  else return ...
+```
+
+Thirdly, it can evaluate to a returned (successful) thread in all cases, containing a variant indicating success or failure:
+
+```
+let load x = if ...
+  then return (`Error (Permission_denied "Cannot read ..."))
+  else return (`Ok ...)
+```
+
+The first two are very similar, as if the exception raised in the first example reaches Lwt it will be turned into a failed thread.
+The difference is that, depending on exactly how the threads are scheduled, a raised exception might make it back to the caller as an exception or as a failed thread. A buggy caller might fail to handle one or the other case (though using a `try_lwt` block will handle both correctly).
+Therefore, using the second form instead of the first is preferred.
+
+The Mirage interfaces generally prefer the third option.
+Most of the module types in `V1` include an `error` type. Some leave it abstract, while others define it in various ways:
+
+- `RANDOM` and `FLOW` leave it abstract.
+- `ENTROPY`, `CONSOLE`, `BLOCK`, `NETWORK`, `ETHIF`, `IP`, `UDP`, `TCP` and `STACKV4` define it as a polymorphic variant.
+- `KV_RO` defines it as a regular variant.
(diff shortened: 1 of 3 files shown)


amirmc:
https://github.com/talex5/mirage-www/blob/errors/tmpl/wiki/error-handling.md

(clicky link to view as a markdown file)

jonludlam:
I worry a little about saying that it's OK for any function to raise an exception. For one thing, anything that returns an Lwt thread that also raises an exception is dangerous - if uncaught the exception will bubble right up to the top, and also you lose any semblance of a backtrace. Is there a way around this except for being very careful around every bind?

tmpl/wiki/error-handling.md
```diff
@@ -0,0 +1,384 @@
...
+Looking back at the three uses above:
+
+- For providing human readable errors, any system except 1 will do.
+- Allowing programs to detect and handle certain errors specially works best with systems 1 and 4.
+  System 2 only works if callers require a particular concrete implementation, breaking the abstraction.
+  System 3 provides (I think) no easy way to match on the type of error.
+- For signalling errors in order to trigger rollback or freeing of resources, any system will do.
+
+It is perhaps worth noting that handling specific error cases specially is extremely rare.
+I don't think I've seen any examples of this feature being used with the `Error` variants in the Mirage code-base.
```
avsm: There are a few cases where you want to handle specific error cases.  Consider the difference between a temporary network transmit failure (e.g. Netif being out of buffer space) vs a connection refused being signalled from the remote host.   In this case though, catching an `EAGAIN` or equivalent exception would be reasonable.

tmpl/wiki/error-handling.md
```diff
@@ -0,0 +1,384 @@
...
+It may be useful to indicate when an error is safe to display to an untrusted client vs one only the admin should see.
+"System too busy; try again later" is an example.
+Normally, however, any unknown exceptions should be logged and an untrusted client given a generic "server failed" message.
+It would be useful to include an opaque tag in such messages so that the error could be correlated back to the log entry if needed.
+
+In a truly secure system, error details should be hidden not only from remote users but also from other components of the program.
+This can be done using "sealed exceptions", whose payload can be read only by the logging system.
+
+Sometimes it's useful to organise errors into a hierarchy, allowing callers to catch e.g. `Exception`, `Network_error`, `TCP_error`, `TCP_connection_refused`, etc.
+I don't see any way to do this in OCaml.
```
avsm: This could be done via open types in OCaml 4.02.0+ (which in fact use the exception memory representation)

avsm:
One way around the Lwt "bare exception" problem would be to establish an Async-style monitor model.  The Lwt scheduler would then be aware of the current monitor, and be able to dispatch a bubbled up exception to the nearest one.

tmpl/wiki/error-handling.md
talex5: - If the transmit buffer is full, we just wait until there's space.
- I'd say `EAGAIN` is a perfect example of something that should *not* be an exception. It's something you probably want to handle, and will probably forget unless prompted. Let's not have any in our APIs ;-)
- I'd be tempted to say that connection refused should be a return code too (`Network_error of exn`), since unreliable networks are something else you might often want to handle. The payload should be an exception so people who don't care can just throw it.

tmpl/wiki/error-handling.md
talex5: Not sure how that helps. My point (not clearly expressed) is that catching e.g. `Network_error` here should also catch the more specific `TCP_error` and `TCP_connection_refused`.

talex5:
@jonludlam "I worry a little about saying that it's OK for any function to raise an exception.". This was more intended as a "This is how OCaml works. We need to deal with it." Out-of-memory is the canonical example of an exception you have to live with.

Not sure what you mean about exceptions bubbling up to the top. For a server, an exception would normally just abort that single client request but continue serving. An exception reaching the top-level would normally only happen during boot, but that's probably what you want.

tmpl/wiki/error-handling.md
```diff
@@ -0,0 +1,384 @@
...
+the information reported to the user is just a small number stored in `errno` and
+errors are easily ignored accidentally.
+
+With Mirage, we have the opportunity to provide something much better.
+OCaml's structured types and exceptions can provide rich diagnostic information that is readable to humans and to machines,
+and its exhaustiveness checks mean we can force callers to consider errors where necessary.
+
+There are three main aspects to consider:
+
+- Providing diagnostic information to humans.
```
mor1: worth distinguishing whether this is to be consumed at the time (standard kind of debugging backtrace) vs post hoc analysis from logs?

tmpl/wiki/error-handling.md
```diff
@@ -0,0 +1,384 @@
...
+
+  val read:
+    t -> int64 -> page_aligned_buffer list ->
+    [ `Error of error | `Ok of unit ] io
+
+  ...
+```
+
+The implication is that callers of `read` should write code to handle each case.
+But how?
```
dbuenzli: Not really. You could have another binding function (&>>=) that would let you handle only the `Ok case and propagate the error up without having you to bother about it except at the top level.

tmpl/wiki/error-handling.md
```diff
@@ -0,0 +1,384 @@
...
+  type error = Unknown_key of string
+
+  val read:
+    t -> string -> int -> int ->
+    [ `Ok of page_aligned_buffer list | `Error of error ] io
+```
+
+For a simple in-memory store this error type allows us to represent all possible errors.
+However, if we want a store backed by a block device, there is no way to pass through errors such as the `Disconnected` one above.
+It is usually impossible to declare all the possible errors that an implementation of a generic interface might produce.
```
dbuenzli: That seems to be definitional defect that should be corrected. If you had a proper error type `('a, 'b) error = [````Ok of 'a | ````Error of 'b]`  and `Unknown_key` was a variant you could get through this.

tmpl/wiki/error-handling.md
```diff
@@ -0,0 +1,501 @@
...
+    | `Disconnected of disconnected
+  ]
+
+  type write_error = private [> read_error |
+    | `Is_read_only of is_read_only
+  ]
+
+  type error = private [> read_error | write_error ]
+
+  val pp_error : formatter -> error -> string
```
samoht: should be `formatter -> error -> unit`

talex5: Fixed.

hannesm:
I guess this should be updated, now that there is some implementation and agreement (https://github.com/mirage/mirage-dev/pull/107)

yomimono:
I think it would be nice to revisit this now that we have `result` and `rresult`.

talex5:
Yes, it does need updating. It's on my list of things to sort out eventually...

talex5:
Closing this because Mirage adopted a different system in the end.

PR #283, opened 2015-02-11
[Testing] Use Mini-OS 0.6 for stack-protector fix

talex5:
Testing with new mirage-xen-libs.git to see if that fixes the problem.

### DIFF
--- a/.travis-mirage.sh
+++ b/.travis-mirage.sh
@@ -1,0 +1,57 @@
+## mirage setup
+
+set -ex
+
+## fetch+execute the OCaml/opam setup script
+wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/master/.travis-ocaml.sh
+sh .travis-ocaml.sh
+
+## install mirage
+export OPAMYES=1
+eval $(opam config env)
+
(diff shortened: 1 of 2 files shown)


talex5:
I was only intending this PR to be used for testing. It should be reverted when the new mirage-xen-minios appears in opam-repository (or if @mor1 modifies his script to pin the new version).

avsm:
I have a strong preference for the Travis deployment script to be stored in this repository rather than @mor1's branch which noone else can modify.  Once it moves into a `mirage/ocaml-travis-ci-skeleton` or similar repository, it can go back to be being sourced.

mor1:
hm-- the scripts should be being sourced from `ocaml/ocaml-travisci-skeleton` now that @samoht merged them. i thought i'd already fixed that -- will do so now.

mor1:
(or would you prefer to use a `mirage` fork of that `ocaml/...` repo?)

avsm:
That repository on `ocaml/` is fine for now, but I'd really like to see some minimal documentation in the README explaining how to use the deployment feature.

mor1:
the README in this repo or in the travis skeleton scripts repo?

can add that easily enough either way.

avsm:
the travis scripts -- the place where the shell scripts actually exist, so that other people can use them.

PR #297, opened 2015-03-19
Add mirage-vnetif

MagnusS:


avsm:
thanks! +jitsu as well? (should tag a release of that also)

PR #313, opened 2015-04-29
Tweak the stats reporting a little bit

samoht:


samoht:
This should hopefully trigger a recompilation with the latest version of mirage tcpip which fixes the memory leak.

PR #317, opened 2015-05-06
Add an option to serve mirage-www over HTTPS

samoht:
Part of the logic should be migrated into the mirage tool. Also the website has numbers of external http links (for instance the fonts) which won't load properly because of the non-mixed contents policy of most browsers .

@avsm I'll let you decide if you want to merge that one or just use it for your demo.


avsm:
Great thanks! Not for merge just yet

samoht:
I've forgotten to run the stats thread in the first batch of commits. Fixed now. Should also probably do something for people trying to connect to port 80 (redirection?)

hannesm:
maybe add HSTS headers in a reply to port 80 (https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security)?

samoht:
https://www.owasp.org/index.php/HTTP_Strict_Transport_Security

> For HTTP sites on the same domain it is not recommended to add a HSTS header but to do a permanent redirect (301 status code) to the HTTPS site.

so serviing a static redirect page might be enough.

avsm:
Redirection is a good idea...

hannesm:
agreed

samoht:
Redirection added

samoht:
@avsm I've added the redirection of port 80 to 443. ready to be deployed! :-)

amirmc:
:+1:

samoht:
I've rebased against master and I've added hooks to be able to change the domain name a bit more easily. Now the domain and the scheme can be set at compile time:

```
$ TLS=1 NAME=mirage.io mirage configure --xen && make
$ TLS=0 NAME=openmirage.org mirage configure --xen && make
```

Only these two names are supported from now. I guess the first deployment should be checked a little bit to be sure there are no typo errors (ie. the HTTP -> HTTPS redirection works, the absolute links are not broken, etc), appart from that it should hopefully work fine.

samoht:
@avsm need to sync-up to integrate https://mirage.io deployment with that PR. Normally it's just a matter of adding the right runes in `.travis-ci.yml`.

samoht:
After lots of cleanup I'm almost happy with the result. You can now set-up static redirections and choose the hostname at configuration time. All the `.ml` files have a well-documented `.mli`. I found some inconstancies in the `Cowabloga` API that I will probably upstream -- actually most of the libraries can/shoud be upstreamed somewhere.

samoht:
With the latest commit, I've added 2 deploy targets. I'm not sure how this is supposed to work, I need to read to deploy scripts. @mor1 can you check that what I wrote makes sense (at least the `.travis.yml`)? Not sure exactly where the deployed images are pulled and restarted ...

PR #326, opened 2015-06-03
Update TROVE - add mirage-seal

amirmc:


samoht:
Need some gymnastics to make Travis happy. This is done on my tls branch (which also contains some other stuff unfortunately)

PR #340, opened 2015-06-18
Add breaking-changes template

talex5:
See: http://lists.xenproject.org/archives/html/mirageos-devel/2015-06/msg00133.html


samoht:
Thanks!

PR #343, opened 2015-06-26
Fix initialisation of stats to be init only once (and not once per re…

samoht:
…quest ...)

### DIFF
--- a/src/dispatch.ml
+++ b/src/dispatch.ml
@@ -204,10 +204,10 @@ module Make_localhost
       log c "conn %s closed" cid
     in
     log c "Listening on %s" (Site_config.base_uri domain);
-    Stats.start ~sleep:OS.Time.sleep;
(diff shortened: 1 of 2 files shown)


samoht:
This was quite bad as it added a new stats handler on every new request ... which is causing a exponential number of calls to `GC.stats` ...

djs55:
:( I'm not sure if my pretty graph thing does a good job with exponentials.
Maybe I need to log something ;)

On Fri, Jun 26, 2015 at 11:50 AM, Thomas Gazagnaire <
notifications@github.com> wrote:

> This was quite bad as it added a new stats handler on every new request
> ... which is causing a exponential number of calls to GC.stats ...
>
> —
> <https://github.com/mirage/mirage-www/pull/343#issuecomment-115642801>.
>



-- 
Dave Scott

PR #349, opened 2015-06-28
Load the stats with the same protocol as the main page

djs55:
It's illegal for .js loaded over https to load XML over http.
We should use the same protocol for all requests.

Signed-off-by: David Scott <dave@recoil.org>


Drup:
Still not working. :(

PR #354, opened 2015-06-30
Enable tracing

talex5:
This might help us find out why the web-site sometimes stops. Ensure it is deployed with `on_poweroff = 'preserve'` in the `.xl` config file so we can still read the trace if the domain crashes.

To dump the trace, run:

```
mirage-trace-viewer -d www -w trace.ctf
```

There are pre-compiled binaries of mtv here: http://talex5.github.io/mirage-trace-viewer/mtv.xml (hopefully these will run from dom0, unless it's a really old OS).

It would be a good idea to look at @djs55's stats after deploying this to check the overhead isn't too big.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,18 +6,19 @@ env:
   - OCAML_VERSION=4.02 MIRAGE_BACKEND=unix FLAGS="-vv --net socket"
   - OCAML_VERSION=4.02 MIRAGE_BACKEND=unix FLAGS="-vv --net direct"
   - OCAML_VERSION=4.02 MIRAGE_BACKEND=xen  DEPLOY=1 XENIMG="openmirage.org"
-    PINS="mirage-net-xen:1.4.2"
+    PINS="mirage-net-xen:1.4.2 lwt:https://github.com/talex5/lwt.git#tracing"
     FLAGS="-vv --net direct --dhcp false --ip 46.43.42.147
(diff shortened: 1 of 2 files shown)


samoht:
I think that's a great idea. That would be even better if we can generate a webpage to see the live tracing (eg. `/stats/tracing` or something similar). In the longer term, we should persist all the stats elsewhere and provide the "tracing page" as a reusable component for other mirage apps.

talex5:
How about a cron job that does `mirage-trace-viewer -d www --html trace` once a minute, then serve up the `trace` directory somewhere? Better to do the tracing outside of the unikernel, otherwise the act of serving up the trace will add to the trace... also it won't work if it crashes.

talex5:
(note: access to the trace page should be restricted, since the trace may include private information such as the IP addresses of visitors)

djs55:
@talex5 looks like this needs a rebase + conflict resolution. Were there any downsides to this (apart from worries about overhead?)

talex5:
I've updated it (and also updated lwt tracing for 4.03). Not sure what the performance effect will be...

djs55:
Needs a rebase again, sorry :( Once it's rebased and the CI passes I think we should merge.

talex5:
Closing as tracing no longer works with recent Lwt versions.

PR #355, opened 2015-06-30
Avoid loading and parsing all the templates on every request

samoht:


avsm:
we need a strong dose of @yallop's staging here :-)

samoht:
as expected, this gives a big boost in performance to load the webpage ...

Issue #338, opened 2015-06-17
Misleading flag in Makefile (IPADDR ?= dhcp)


djs55:
The Makefile suggests that you should configure this variable:
```
IPADDR ?= static
```
but actually the `config.ml` does something different:
```
     let staticip =
        let address = Sys.getenv "ADDR" |> Ipaddr.V4.of_string_exn in
        let netmask = Sys.getenv "MASK" |> Ipaddr.V4.of_string_exn in
        let gateways =
          Sys.getenv "GWS" |> split ':' |> List.map Ipaddr.V4.of_string_exn
        in
        { address; netmask; gateways }
      in
      direct_stackv4_with_static_ipv4 console tap0 staticip

    | false ->
      let net =
        try match Sys.getenv "NET" with
          | "socket" -> `Socket
          | _        -> `Direct
        with Not_found -> `Direct
      in
      let dhcp =
        try match Sys.getenv "DHCP" with
          | "1" | "true" | "yes" -> true
          | _  -> false
        with Not_found -> false
      in
```
(I think `NET` is ok but the others are wrong or missing)

yomimono:
I went to patch this as a side effect of building mirage-www locally and got distracted by new variables REDIRECT and HOST; is there already a summary of their semantics somewhere?  If not I can puzzle it out from config and dispatch and write one up. (@samoht?)

samoht:
err no sorry I haven't written anything about them yet. If you have the time to draft that, that would be very helpful! Thanks!

mor1:
This looks to me like it can be closed now -- @djs55 any objections?

djs55:
The doc is much better now, thanks!

PR #356, opened 2015-06-30
Makefile: more documentation for environment variables

yomimono:
Per #338 , update and expand the documentation for environment variables in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ HOST   ?= localhost
 # to redirect all http requests to https://somewhereelse.org:
 # REDIRECT = https://somewhereelse.org 
 # (redirect must be an http:// or https:// url)
-# if https is being served and REDIRECT is not specified, it will be assumed to be https://$HOST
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,13 +16,37 @@
...
+# default static settings will be used.
+
+# to serve pages over https:
+# TLS = true
+# to make it work with FS=crunch, put the certificate in tls/tls/server.key and the private key in tls/tls/server.pem
+
+# to redirect all http requests to https://somewhereelse.org:
+# REDIRECT = https://somewhereelse.org 
+# (redirect must be an http:// or https:// url)
+# if https is being served and REDIRECT is not specified, it will be assumed to be https://$HOST
```
samoht: also if TLS is set, `http://$HOST/<path>` will be redirected to `https://$HOST/<path>`

samoht:
That's very helpful, thanks! I've restarted the failing job (because I like green crosses) and I'll merge shortly after that.

Makefile
yomimono: If I understand correctly that's what line 49 says, but I'll reword the comment to make it a bit clearer what I mean and then you can tell me whether I'm correct...

PR #369, opened 2015-07-13
Add c3 library dependency

mor1:


mor1:
Travis is green, merging.

PR #388, opened 2015-09-09
Add notes from weekly call on 2015-09-09

amirmc:
Will merge once travis is green.

If anyone wants to take a look in the meantime, please do so!


amirmc:
Travis is green. Merging.

PR #389, opened 2015-09-14
[not for merge] Mirage-www, functoria edition.

Drup:
This should not be merged, the deployment scripts are not updated.
I did this to "try it out" (and see the pretty graph at the end, really).

The commit first two commits fix compat. The rest is just for goodies and self-documentation:

Here is the graph visible by the user with mirage describe before conversion to functoria:
Mirage-www before.

You can clearly see the network stack on the right bottom and the user functor on the top.

Then I used the combinators now available to rewrite the config file and to improve the code in general.
Here are the help pages for the [configure](https://gist.github.com/Drup/fca9a172ca7123eafaff#file-configure-help-page-for-mirage-www) and for the [actual unikernel](https://gist.github.com/Drup/fca9a172ca7123eafaff#file-runtime-help-page-for-mirage-www)
I should probably merge the various kv_ro option.

Here is the dot output for the whole mirage-www:
Mirage-www after

This is quite big. The generic network stack (dhcp + static + socket) is on the bottom left, key/value stores are on the right and you can see the two dispatch modules in the middle.

It is possible to cut down the graph by instanciating some of the options. Here is the graph full evaluated with the *socket* stack and tls *enabled*:

Mirage-www socket and tls

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,6 +1,5 @@
 PKG cohttp cow cowabloga cstruct lwt mirage mirage-unix mirage-http
 PKG mirage-net-unix mirage-clock-unix mirage-console.unix mirage-fs-unix
 
-B src/_build/
-B src/build/**
+B src/_build/**
 S src/
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,17 @@ install: wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/ma
 script: bash -ex .travis-mirage.sh
 env:
   matrix:
-  - OCAML_VERSION=4.02 FLAGS=-vv MIRAGE_BACKEND=unix MIRAGE_NET=socket
-  - OCAML_VERSION=4.02 FLAGS=-vv MIRAGE_BACKEND=unix MIRAGE_NET=direct
-  - OCAML_VERSION=4.02 FLAGS=-vv MIRAGE_BACKEND=xen  DEPLOY=1 XENIMG="openmirage.org"
-    IP="46.43.42.147" NETMASK="255.255.255.128" GATEWAYS="46.43.42.129"
-    TLS=0 HOST="openmirage.org" REDIRECT="https://mirage.io"
+  - OCAML_VERSION=4.02 MIRAGE_BACKEND=unix FLAGS="-vv --net socket"
+  - OCAML_VERSION=4.02 MIRAGE_BACKEND=unix FLAGS="-vv --net direct"
+  - OCAML_VERSION=4.02 MIRAGE_BACKEND=xen  DEPLOY=1 XENIMG="openmirage.org"
+    FLAGS="-vv --net direct --dhcp false --ip 46.43.42.147 \
+    --netmask 255.255.255.128 --gateways 46.43.42.129 --tls false \
+    --host openmirage.org --redirect https://mirage.io"
     UPDATE_GCC_BINUTILS=1
-  - OCAML_VERSION=4.02 FLAGS=-vv MIRAGE_BACKEND=xen DEPLOY=1 XENIMG="mirage.io"
-    IP="46.43.42.146" NETMASK="255.255.255.128" GATEWAYS="46.43.42.129"
-    TLS=1 HOST="mirage.io" SECRETS=fat
+  - OCAML_VERSION=4.02 MIRAGE_BACKEND=xen DEPLOY=1 XENIMG="mirage.io"
+    FLAGS="-vv --net direct --dhcp false --ip 46.43.42.146 \
+    --netmask 255.255.255.128 --gateways 46.43.42.129 --tls true \
+    --secrets-kv_ro fat --host mirage.io"
     UPDATE_GCC_BINUTILS=1
   global:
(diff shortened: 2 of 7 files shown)


amirmc:
I edited the title (just in case anyone gets trigger-happy).

samoht:
nice graphs!

samoht:
why do the `if` not put at the higher-level in the large graph (they seem to appear under `$`)?

Drup:
@samoht I don't do this transformation anymore. It was causing more harm than good and it was making the graph a big mess. I needed it for bad reasons that are not valid now that the engine is properly graph based.

mor1:
(idle request: I presume the `.dot` output is captured to a file
separately?  this would be really nice to update the very old, incorrect,
hand-generated module dependency graph in the various decks that we keep
using, but i'd want to tweak the dot file when doing that.)

On 14 September 2015 at 19:02, Gabriel Radanne <notifications@github.com>
wrote:

> @samoht <https://github.com/samoht> I don't do this transformation
> anymore. It was causing more harm than good and it was making the graph a
> big mess. I needed it for bad reasons that are not valid now that the
> engine is properly graph based.
>
> —
> <https://github.com/mirage/mirage-www/pull/389#issuecomment-140161560>.
>



-- 
Richard Mortier
mort@cantab.net

Drup:
@mor1 : By default, it shows the dot directly (with `xdot`) and there is an option `-o` to output to a file.

samoht:
> @samoht I don't do this transformation anymore. 

ok.

Also in the man page, the options which can be be enable because other options are enabled do not seem to be shown. If that information available somewhere? Does the describe command can show that tree of keys?

Drup:
In the manpage, I show 
- The switching keys
- The keys of the current evaluated tree. 

Indeed, I don't add the relative dependencies in the manpage. There is no way to show a tree in those, really, so I concentrated on having a good dot output.

samoht:
Having the tree of keys somewhere (not necessary in the man page) is still useful I think. The dot output is nice but it carries a lot of information. Projecting the graph into something smaller (e.g. only the "if" nodes) would be useful at one point (but it is neither blocking integration or urgent). Not sure where to report this though :-) should it be on Functoria bug tracker?

Drup:
> Projecting the graph into something smaller (e.g. only the "if" nodes) would be useful at one point

Ok

> should it be on Functoria bug tracker?

Yes

dbuenzli:
> Having the tree of keys somewhere (not necessary in the man page)

If you are still interested in having it in the man page https://github.com/dbuenzli/cmdliner/commit/ce87b2495789677d5bca136d59b3a7bdc9f25ef4 may help.

Drup:
Updated. what is the `XENIMG` variable for ? It only changes the name of the unikernel ... which is not really used anywhere.

Drup:
Hum, the "PR" variable doesn't seem to be set (and travis doesn't set it). What is it for ?

mor1:
@drup `XEMIMG` is used by the Travis build scripts to create the images that are committed back to a deployment repo. See eg, https://github.com/ocaml/ocaml-travisci-skeleton/blob/master/travis_mirage.ml. Not sure about `PR` though... (`TRAVIS_PULL_REQUEST` seems to be the var that's set by Travis.)

Drup:
@mor1 It seems `XENIMG` is rewritten after a while ... Can we just assume the name is constant (`www`) and avoid the additional variable in `.travis.yml` ?

Drup:
Also: Why is the script in another repo ? It makes tinkering with it difficult.

samoht:
These scripts are also used by `mirage-decks` and `mirage-tutorials` I think

mor1:
@samoht yes -- used by decks and tutorials (@drup hence can't assume `www`)

samoht:
Superseded by #410

Drup:
@samoht not really. This actually converts the website to the new API. #410 is just the minimal compatibility change.

samoht:
Hum right. Can you try to rebase to master? Thanks!

Drup:
@samoht I will do that tomorrow, but it needs more work anyway, to make sure that the travis scripts are correct (and I can't really test them ...).

samoht:
I'm working on the Travis scripts currently, so just'd be just a rebase of the OCaml code to use the new API. Thanks!

Drup:
@samoht done. What, I think, remains:

- [ ] Figure out the name of the image
- [ ] Decide what to do with the PR key (it's not used, see previous discussion)
- [ ] Check that the deployment actually works

samoht:
Thanks! Removing my patch seems to have worked fine, yay!

If I remember correctly, the XENIMG var is used by the deployment script to name the VM so we should keep it. What `PR` are you talking about? Is that `TRAVIS_PULL_REQUEST`? If yes, that's set by TravisCI and it is used to avoid deploying images while testing locally.

samoht:
Hum actually I'm not sure it worked, I'll debug that...

Drup:
> What PR are you talking about? Is that TRAVIS_PULL_REQUEST? If yes, that's set by TravisCI and it is used to avoid deploying images while testing locally.

No, Look at the current `config.ml`, the one in this pr and at https://github.com/mirage/mirage-www/pull/389#issuecomment-152812088. There is something fishy.

Drup:
Oh, apparently it was fixed! It's using the proper env variable now we also need to fix that before merging.

samoht:
Hum weird, I had to ping the VM from dom0 to unfreeze it, and now the site is live. @avsm or @yomimono do you have an idea where this could come from?

Drup:
@samoht I fixed two of the three issues, the remaining one is all yours. ;)

PR #395, opened 2015-10-06
added installation instructions for arch linux

matildah:


avsm:
thanks!

tmpl/wiki/install.md
```diff
@@ -59,6 +63,8 @@ Finally, install the MirageOS command-line tool.
     $ opam install mirage
     $ mirage --help
 
+If you're using Arch Linux and `opam install mirage` gives errors about not finding camlp4 and camlp4 was installed with pacman (and not by OPAM), then running `opam install mirage` without having executed the `eval` line lets `opam` find the system-wide installation of camlp4.
+
```
Drup: But, hum, that's not a very good advice, we should report that bug in opam instead ....

PR #398, opened 2015-10-15
Wrap video in flex-video class

amirmc:
This is part of Zurb Foundation and allows the video to scale
depending on the device it's being viewed on (i.e. in this case
it improves how the post looks on a mobile device).


amirmc:
I'm skipping the CI for this one.

PR #400, opened 2015-10-23
Fix typo.

schmidh:
Fix typo also in accordance to mirage-skeleton.


talex5:
Thanks - some Travis builds failed for other reasons, but the patch is obviously OK so merging anyway.

PR #402, opened 2015-11-01
Minimal changes to fix travis

Drup:
It seems that CI pins @samoht's mirage repository, which now uses functoria, which breaks things. This fixes it.

The more involved changes in #389 are superfluous, so I made another PR.


Drup:
Hum, I think the nocrypto detection is wrong somehow.

Drup:
All green!

Drup:
Ping ? @samoht ?

PR #411, opened 2015-11-19
Updated install.md to reflect the need for OPAM 1.2 and to emphasis f…

trcollinson:
…and to emphasis further that system installs may be too old.


mor1:
Thanks!

PR #423, opened 2016-01-01
add short note about hackathon in marrakech

hannesm:


hannesm:
@amirmc @avsm @mor1 would be great to have this one merged asap, thx!

amirmc:
lgtm! Thanks for this :)

hannesm:
looks like travis committed the unikernels, but they fail to start up... (mirage.io is offline!)

amirmc:
Last time this happened was due to a failed git pull (that didn't affect the actual live site though. Not sure what's up with that). 

@mor1, could you take a look when you get the chance? 

Best wishes,
Amir

--
Sent via mobile

mor1:
```
Fatal error: exception Failure("/blog/hackathon-marrakech2016.md not found")
```
...which is causing the domain to crash out, and then the `.xl` restarts it.

the file is `hackathon.md` by the looks of it.

mor1:
have submitted and merged a PR to fix the filename.
give it 10min for Travis to catch up...
(perhaps best to check the site works, not just compiles, locally before merging next time? site currently offline as a result of this.)

mor1:
site now back up and post live

amirmc:
Thank you, Mort! 

Unfortunately, it's not really feasible to check that every PR runs locally before merging. We should encourage authors to do this and try to make it easy for them (I thought that @hannesm had already done this -- or that an error would lead to a CI failure). 

Is there a way to make the site fail more gracefully for a situation like this?  Feels like such an error slipping in shouldn't be able to take down the site. 

Best wishes,
Amir

--
Sent via mobile

hannesm:
thanks @mor1 for fixing.  and sorry for not extensively testing (I somehow had the impression such errors are caught at compile time, and forgot to rename the file)...

PR #428, opened 2016-01-05
add links to devel.unikernel.org

amirmc:
We can use [devel.unikernel.org](http://devel.unikernel.org) for any discussions on blog posts. I've added links to the two recent posts.

We've not had any comment functionality on the site so this can fill that gap.

### DIFF
--- a/tmpl/blog/hackathon-marrakech2016.md
+++ b/tmpl/blog/hackathon-marrakech2016.md
@@ -1,3 +1,7 @@
 The first MirageOS hackathon will take place in Marrakech, Morocco, from 11th till 16th March 2016.  It is open for everybody.  The main goal is to get together people motivated to contribute to MirageOS.
 
 Find more details on [the hackathon website](http://marrakech2016.mirage.io).
+
+*Edit: discuss this post on [devel.unikernel.org][discuss]*
+
+[discuss]: http://devel.unikernel.org/t/1st-mirageos-hackathon/24/1
--- a/tmpl/blog/hackathon-marrakech2016.md
+++ b/tmpl/blog/hackathon-marrakech2016.md
@@ -4,4 +4,4 @@ Find more details on [the hackathon website](http://marrakech2016.mirage.io).
 
 *Edit: discuss this post on [devel.unikernel.org][discuss]*
 
-[discuss]: http://devel.unikernel.org/t/1st-mirageos-hackathon/24/1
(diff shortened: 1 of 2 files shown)


hannesm:
hmm, this adds another channel, next to the mailing list and IRC, for input, which needs to be checked!?

I'm not really convinced that we need more communication channels, esp since the ones in place are not used too much (there are every now and then no people answering questions on IRC).  What is the value of discussing e.g. the hackathon blog post (I'd prefer people to send a mail to the mail address mentioned on marrakech2016.mirage.io instead of commenting somewhere on the internet which might get lost).

but, if you all agree that this is a good idea, so be it.  I personally am not a big fan of forum or blog comments.

amirmc:
I do see your point but there are already a lot of channels to keep up with. Off the top of my head, I usually try to keep an eye on GitHub (*so many repos now*), the mailing list, Twitter, Hacker News and Reddit — I’m not a fan of IRC (I only go there to notify people about the fortnightly call) :)

**About devel.unikernel.org**
[devel.unikernel.org](http://devel.unikernel.org) is trying to help bring together the different projects so that we can collaborate more easily and help onboard new people into the unikernel ecosystem as a whole (of which MirageOS is a big part).  Cross-cutting discussions need to take place somewhere and I was reluctant to create yet another mailing list for this so I considered alternatives (see my thoughts over at [Unikernel-Systems/unikernel.org#/25][] inc. about email interaction).  I believe this is also going to give MirageOS a boost as not everyone likes mailing lists. The previous post about [why we need unikernel.org](https://mirage.io/blog/unikernel-org) also has some details.

**Hackathon comms**
Regarding the hackathon, people *must* email to register their interest/intent in taking part. It's *completely* reasonable to mandate that :) — having said that, I'd much prefer a public place for discussions as other people can also get involved and see what's happening. It's possible to 'watch' that topic just like a GitHub issue and then interact via email (i.e. you won't need to visit the site). IMHO leaving such traces on the Internet are really valuable in demonstrating our momentum (otherwise it's hidden away).

*( sorry for the wall of text — I'm not just writing this for you Hannes ;) )*

[Unikernel-Systems/unikernel.org#/25]: https://github.com/Unikernel-Systems/unikernel.org/issues/25

amirmc:
If there aren't any other questions, I'll be merging this in the next few hours (I'll likely need to update `http` -> `https` beforehand)

PR #433, opened 2016-02-15
Update compiler requirements

ansiwen:
Mirage OS also supports the latests ocaml compilers. The current installation guide suggests to install 4.01.0, but in fact this is a lower bound.


yomimono:
Thanks for the PR! This looks great. :)

PR #441, opened 2016-03-15
Add rough versions of notes from calls

amirmc:
I spent most of today sorting through the outstanding notes for the calls.

These are just in bullet point form and it's obvious that they're somewhat rough.  If you spot typos or errors, I think it may be better to fix them post-merge (since you can push directly to `master` then).  In any case, I suspect people will be happier with rough notes than no notes :)

Thanks to @GemmaG for also taking notes during the last two calls. They've been folded into the set as part of this PR.


amirmc:
Hmmm... not sure why I can see commits from December here. They don't seem to be reflected in the actual file changes.  Let me know if it's a problem.

samoht:
Great!!! thx for taking the time to put these notes online, that's very useful.

PR #449, opened 2016-05-04
tweaks

hannesm:
@samoht complained about the markdown... I guess having a pure signed text version and a separate markdown version actually using markdown is better


samoht:
Thanks!

PR #450, opened 2016-05-04
Add hackathon trip reports from many wonderful authors!

yomimono:
Authors are credited in the sections they contributed.

Signed-off-by: Mindy Preston <mindy.preston@docker.com>


pqwy:
It might be cool to move the linked-to images locally; it would make the text more future-proof. For example, Imgur expires content after no-one has accessed it for 6 months. En-dash.

yomimono:
Good point.  I've made an update with local copies of the imgur photos -- do you want yours mirrored too, @pqwy ?

avsm:
Is there _any_ repo that @yomimono won't sneak a cat picture into??

tmpl/blog/2016-spring-hackathon.md
```diff
@@ -0,0 +1,231 @@
+# Hackathon Trip Reports
+
+We're looking forward to the next MirageOS hackathon already!  We've collected some reports from those who were present.  Thanks to the folks who put in the time and effort to organize the event and our wonderful hosts, and a huge thanks to everyone who documented their hackathon experience!
```
avsm: Could we link to the original hackathon page here? We never actually blogged about it here, I just realised.

hannesm: we did https://mirage.io/blog/hackathon-marrakech2016

hannesm:
@yomimono yes, please add those I stole from the internet and gave to @pqwy to use... they might disappear at any point in time..

tmpl/blog/2016-spring-hackathon.md
avsm: my bad!

yomimono: I've added a link to the marrakech2016.mirage.io site, since the announcement blog post was pretty much a shell around that href anyway.

hannesm: hmm, ok... I had the hope I could silently destroy the marrakech2016.mirage.io unikernel now that the hackathon is over... ;)

hannesm: but no worries, I'll preserve the service then

avsm:
thank you all for the trip reports!

PR #452, opened 2016-05-05
minor formatting

hannesm:


hannesm:
would be nice if someone would merge this

PR #454, opened 2016-05-06
Minor fixes and clarifications to Hello World and Mirage WWW doc pages.

mquinn:
The examples using netcat assume the user is using netcat-openbsd
rather than netcat-traditional; the former allows the -4 flag to
force IPv4 address use only, the latter does not support the flag.
The -4 flag is not needed here and thus has been removed
from all examples. The various flags have been consolidated as well.

The two sections regarding use of Mirage with static IP addresses assumes
the user has an existing tap0 interface, which may not be the case.
Additionally, on both pages, the command given to bring tap0 up does not work on
recent versions of ifconfig, and on the mirage-www page, is missing the
"up" argument at the end. To remedy this, statements explaining
how to set up tap0 have been added, and the commands to bring tap0
up have been updated/corrected and moved up to ensure they are executed prior to
the start of the unikernel.

### DIFF
--- a/tmpl/wiki/hello-world.md
+++ b/tmpl/wiki/hello-world.md
@@ -712,9 +712,9 @@ the commonly available _netcat_ `nc(1)` utility. From a different console
 execute:
 
 ```
-$ echo -n hello udp world | nc -n4w1 -u 127.0.0.1 53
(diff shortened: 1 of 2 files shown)


djs55:
Thanks!

PR #457, opened 2016-05-20
Add --http-port and --https-port options

talex5:
Makes it easy to test without having to run as root.


yomimono:
Excellent!  Thanks for this.

PR #458, opened 2016-05-20
Update Lwt tutorial

talex5:
Main changes:

- Added a slightly longer introduction about the `Lwt.t` type and how
  threaded code corresponds to synchronous code. Hopefully this is
  clearer.
- Removed the `<&>` and `<?>` operators, since we hardly use them.
- Changed `bind` to `>>=`, since that's how we use it.
- Removed all mention of syntax extensions, since we no longer use them.
- Moved discussion to the event loop nearer the top, since it's quite
  important.
- Added discussion of `Lwt.async`, since that's pretty important and
  people often get it wrong.
- Added discussion of handling errors (results and exceptions), as
  requested.
- Moved echo server example up, since it's very simple.
- Added discussion of `Lwt.wait` as that's how you connect Lwt to
  non-Lwt things.
- Removed all mention of mvars. These are rarely needed, and probably
  over-used by beginners. The example could be handled much more simply
  with `Lwt_list`, which it now links to instead.
- Removed stream server example which followed on from the mvar one.
  Less sure about this, but it seemed unnecessary in an introduction
  and had very little to do with Mirage.
- Added warning about the dangers of `Lwt.cancel` and suggested using
  `Lwt_switch` instead.
- Changed explanation of when and why mutexes are needed. The old text
  seemed a bit muddled to me.
- Updated examples to match current code in mirage-skeleton (and also in
  hello-world.md).
- Added a link from `hello-world.md` to the Lwt tutorial.

It's not perfect and some my choices might be contentious, but hopefully
it's better than it was. At some point we may want to make this less
Mirage-specific and contribute something to the Lwt project itself,
turning this into a page of best practices for using Lwt with Mirage.


tmpl/wiki/tutorial-lwt.md
```diff
@@ -1,22 +1,63 @@
...
   val return: 'a -> 'a Lwt.t
 ```
 
-It is used to construct a thread that immediately returns with the provided value. Once the value is wrapped in an Lwt thread, it cannot directly be used (as the thread may not have completed yet). This is where `bind` comes in:
+This is useful if an API requires a thread, but you already happen to know the value.
+Once the value is wrapped in its Lwt thread, it cannot directly be used (as in general a thread may not have terminated yet). This is where the `>>=` operator (pronounced "bind") comes in:
 
 ```
-  val bind: 'a Lwt.t -> ('a -> 'b Lwt.t) -> 'b Lwt.t
+  val ( >>= ): 'a Lwt.t -> ('a -> 'b Lwt.t) -> 'b Lwt.t
```
hannesm: I'm used to no whitespaces between `(`, `)` and `>>=`

tmpl/wiki/tutorial-lwt.md
```diff
@@ -1,22 +1,63 @@
...
 ```
 
-`bind t f` creates a thread which waits for `t` to terminate, then passes the result to `f`. If `t` is a sleeping thread, then `bind t f` will be a sleeping thread too, until `t` terminates. If `t` fails, then the resulting thread will fail with the same exception.
+`t >>= f` creates a thread which first waits for thread `t` to return some value `x`, then behaves as the new thread `f x`. If `t` is a sleeping thread, then `t >>= f` will initially be a sleeping thread too. If `t` fails, then the resulting thread will fail with the same exception.
```
hannesm: maybe mention here that lwt is cooperative, thus `t` isn't interrupted at any time

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
-The `OS` module packs several sub-modules depending of the backend, [unix](https://github.com/mirage/mirage-platform/tree/master/unix/lib)
+The `OS` module packs several sub-modules depending on the backend, [unix](https://github.com/mirage/mirage-platform/tree/master/unix/lib)
 or [xen](https://github.com/mirage/mirage-platform/tree/master/xen/lib).
 
 ###Solution
 
 ```
   open V1_LWT (* provides the CONSOLE signature *)
-  open Lwt    (* provides bind and join *)
+  open Lwt    (* provides >>= and join *)
```
hannesm: wondering whether we should propose in this tutorial to only `open Lwt.Infix`, instead of the big `Lwt`?

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
-
-If locking a data structure is still needed between yield points, the `Lwt_mutex` module provides the necessary functions. To obtain more information on thread switching (and how to prevent it) read the Lwt mailing list archive: [Lwt_stream, thread switch within push function](https://sympa.inria.fr/sympa/arc/ocsigen/2011-09/msg00029.html) which continues [here](https://sympa.inria.fr/sympa/arc/ocsigen/2011-10/msg00001.html).
-
-##Exceptions and Try/Catch
+With Lwt, it is often possible to avoid mutexes altogether! The web server from the [Ocsigen](http://ocsigen.org) project uses only two, and the MirageOS source code none. In usual concurrent systems, mutexes are used to prevent two (or more) threads executing concurrently on a given piece of data. This can happen when a thread is preemptively interrupted and another one starts running. In Lwt, a thread executes serially until it explicitly yields (most commonly via `>>=`); for this reason, Lwt threads are said to be [cooperative](http://en.wikipedia.org/wiki/Cooperative_multitasking#Cooperative_multitasking.2Ftime-sharing).
```
hannesm: as said, I'd move this paragraph further up

hannesm:
this is great, thanks @talex5 (nitpicks inline)

tmpl/wiki/tutorial-lwt.md
talex5: Mainly just following http://ocaml.org/learn/tutorials/guidelines.html#Lexicalconventions

> Systematically surrounding operators with spaces simplify the treatment of infix operators which are no more a complex particular case; in effect, whereas you can write `(+)` without spaces, you evidently cannot write `(*)` since `(*` is read as the beginning of a comment. You must write at least one space as in `( *)`, although an extra space after `*` is definitively preferable if you want to avoid that `*)` could be read, in some contexts, as the end of a comment. All those difficulties are easily avoided if you adopt the simple rule proposed here: keep operator symbols well separated by spaces.
In fact you will quickly feel that this rule is not so difficult to follow: the space bar is the greatest and best situated key of the keyboard, it is the easiest to enter and you cannot miss it!

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
 ```
 
 If you are building for a different MirageOS platform, change the `--unix` switches appropriately (presently either `--xen` or `--unix`).
```
aantron: The "appropriately-presently" is a bit nebulous for someone who doesn't know much about MirageOS, maybe just say if you are building for Xen, change `--unix` to `--xen`?

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
-      )
+      Lwt.join [
+        (Time.sleep 1.0 >>= fun () -> C.log_s c "Heads");
+        (Time.sleep 2.0 >>= fun () -> C.log_s c "Tails")
+      ] >>= fun () ->
+      C.log_s c ("Finished")
   end
 ```
 
 This code is also found in `lwt/src/unikernels.ml` in the `mirage-skeleton` code repository. Build it by setting the TARGET environment variable to `heads1` before running `mirage configure`.
```
aantron: The file and repo name would be more helpful as links.

tmpl/wiki/tutorial-lwt.md
samoht: +1

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
+The core of Lwt is based on an event loop. In "standard" (non-MirageOS) settings,
+this loop is started using the `Lwt_main.run` function. However, when using
+MirageOS, the loop is automatically started by the `main.ml` file autogenerated
+by the `mirage` command-line tool.
 
-Here, we wait for the result of `e1`, bind the result to `x` and continue into `e2`. You can replace this with the more natural `lwt` syntax to act as a "blocking let":
+Because it's based on an event loop, threads are actually very cheap in Lwt
+when compared to preemptive system threads. Sleeping actually registers an
+event that will wake up the associated thread when possible. Depending on the
+backend, the event registering slightly differs.
```
aantron: The last sentence worries me – do I have to worry about this difference as a user? I think it's better just to drop it.

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
+       Log.err (fun f -> f "uncaught exception from listen callback \
+                            while handling frame:@\n%a@\nException: @[%s@]"
+                   S.pp_frame data (Printexc.to_string ex));
+       Lwt.return ()
+    )
+)
 ```
 
-###Challenge
+By the way, the reason `async` and `catch` take functions rather than plain threads is so they can catch OCaml exceptions (not just failed threads). Otherwise, if we had `Lwt.catch (fn data) ...` and `fn data` used `raise` then the exception would occur before `Lwt.catch` ran, and it wouldn't be able to handle it.
```
aantron: This paragraph might be easier to understand if it explained the more general reason: "to have control over when the threads are executed. A thread wrapped in a function won't start until the function is called. This allows `async` and `catch` to set up exception handlers."

That is quite a bit less precise, but I think easier for a beginner to understand. And, it gives the right intuition for other uses of threads wrapped in functions. For example, as I'm sure you know, you might have some function that sets up an environment, then runs a user-provided thread, then cleans up the environment. You wouldn't want the user's thread to potentially run before the environment is ready. Maybe you can mention this here very briefly, ("this trick is also used in other places..."), but I don't think it's necessary.

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
+  | Ok b ->
+  return (Ok (a + b))
+```
 
-This solution and application are found in `lwt/src/unikernels.ml` in the repository. Build with target `timeout1`.
-
-Does your solution match the one given here and always returns after `f` seconds, even when `t` returns within `f` seconds?
-
-This is a good place to introduce a third operation to compose threads: `pick`.
+It is often useful to provide some helpers to handle this pattern (using Lwt threads and result types together) more simply:
```
aantron: Relevant: https://github.com/ocsigen/lwt/pull/247. Nothing to do about it here yet, though :)

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
+let invoke_remote msg =
+  let t, waker = Lwt.wait () in
+  let id = new_id () in
+  on_response_to id (fun resp -> Lwt.wakeup waker resp);
+  send_request id msg;
+  t
 ```
 
-###Challenge
+This is mainly useful when interacting with external processes (as in this example), or libraries that don't support Lwt directly,
```
aantron: That trailing comma... :)

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
-  let t = aux () in
-  m_out
+  (* In this example and all those afterwards, we consider Lwt and OS to be
+     opened *)
+  let timeout delay t =
+    Time.sleep delay >|= fun () -> cancel t
 ```
 
-This code can also be found in `lwt/src/mvar_unikernels.ml` in the `mirage-skeleton` repository.
+Note: this example also shows a new convenience operator provided by Lwt. `>|=` ("map") is similar to `>>=` but automatically wraps the result of the function you provide with `return`. It's used here because `cancel` is synchronous (it doesn't return a thread).
```
aantron: The "Lwt. `>|=`" looks like a failed attempt to write `Lwt.(>|=)` when rendered – perhaps you can change the period to a colon or dash?

aantron:
Looks very good. Nits inline as well.

samoht:
LVGTM

tmpl/wiki/tutorial-lwt.md
```diff
@@ -60,11 +99,9 @@ Add `foo.ml` with the following content and edit it:
   module Main (C : CONSOLE) = struct
     let start c =
       (* the guts go here *)
-      return ()
   end
 ```
 
-
 Assuming you are building for POSIX using native kernel sockets, compile the application by:
```
yomimono: I think "using native kernel sockets" doesn't add anything and potentially confuses the reader.  "building for POSIX" could also probably be reworded to something a bit more accessible, although I'm having trouble thinking of something that's not too wordy.

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
+      C.log_s c ("Finished")
   end
 ```
 
 This code is also found in `lwt/src/unikernels.ml` in the `mirage-skeleton` code repository. Build it by setting the TARGET environment variable to `heads1` before running `mirage configure`.
 
-###Syntax Extensions
 
-Using Lwt does sometimes require significantly restructuring code, and in particular doesn't work with many of the more imperative OCaml control structures such as `for` and `while`. Luckily, Lwt includes a comprehensive [pa_lwt](http://ocsigen.org/lwt/api/Pa_lwt) syntax extension that makes writing threaded code as convenient as vanilla OCaml. MirageOS includes this extension by default, so you can use it anywhere you want.
+##Challenge 2: Echo server
```
yomimono: Now that the echo server isn't demonstrating the syntax extensions, it's not clear to me what it's teaching.  Perhaps it could be removed entirely, or used to demonstrate a different kind of composition?

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
-```
-  e1 >>= fun x -> e2
-```
+The core of Lwt is based on an event loop. In "standard" (non-MirageOS) settings,
+this loop is started using the `Lwt_main.run` function. However, when using
+MirageOS, the loop is automatically started by the `main.ml` file autogenerated
+by the `mirage` command-line tool.
 
-Here, we wait for the result of `e1`, bind the result to `x` and continue into `e2`. You can replace this with the more natural `lwt` syntax to act as a "blocking let":
+Because it's based on an event loop, threads are actually very cheap in Lwt
```
yomimono: `s/actually //` or maybe even `s/actually //g`?

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
+  read_arg () >>*= fun a ->
+  read_arg () >>*= fun b ->
+  ok (a + b)
+```
 
-###Challenge
+###Use raise or fail for bugs
 
-In a typical use of a timeout, if `t` returns before the timeout has expired, one would want the timeout to be cancelled right away. The next challenge is to modify the timeout function to return `Some v` right after `t` returns. Of course if the timeout does expire then it should cancel `t` and return `None`.
+If a bug is detected, you should raise an exception. In threaded code you should use `Lwt.fail`, although Lwt should catch exceptions and turn them into failures automatically if you forget.
```
yomimono: `s/Lwt should/Lwt will/` ?

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
+   | ex -> Lwt.fail ex  (* Pass others on *)
+  )
+```
 
-###Challenge
+###Finalize
 
-Write an echo server, reading from a dummy input generator and, for each line it reads, writing it to the console. The server should never stop listening to the dummy input generator. Here is a basic dummy input generator:
+Depending on how the unikernel is set up, an exception may or may not be fatal.
+In general, if you allocate a resource that won't be automatically freed by the GC then you should use `Lwt.finalize` to ensure it is cleaned up whether the function using it succeeds or not:
```
yomimono: `s/GC/garbage collector/` ?

tmpl/wiki/tutorial-lwt.md
```diff
@@ -74,508 +111,327 @@ Assuming you are building for POSIX using native kernel sockets, compile the app
...
 ```
 
-This is in `lwt/src/mvar_unikernels.ml` in the MirageOS code repository. Build with target `int_server`.
-
-
-##Stream Processing
+This solution and application are found in `lwt/src/unikernels.ml` in the repository. Build with target `timeout1`.
 
-The pipelining challenges of the previous section uses mailboxes provided by the `Lwt_mvar` module. One of the feature of such values is that no more than one message can be deposited in the mailbox. This is made obvious by the return type of `Lwt_mvar.put` being in `Lwt.t`. This provides a natural throttle mechanism to our pipeline. However, in specific cases, such a throttle is superfluous.
+Does your solution match the one given here and always returns after `f` seconds, even when `t` returns within `delay` seconds?
```
yomimono: I'd say "always return" instead of "always returns".

yomimono:
Thanks very much for this badly-needed update.  I agree with @hannesm about the placement of the event loop explanation -- having this information is helpful to have before trying to understand `join` and `choose`, in my opinion.

Other than that and some nits commented on inline, it looks excellent to me.

tmpl/wiki/tutorial-lwt.md
talex5: Yes, I'll change this.

tmpl/wiki/tutorial-lwt.md
talex5: Agreed. This makes no sense.

tmpl/wiki/tutorial-lwt.md
talex5: I'll update it to say it's showing how to write a loop with Lwt.

talex5:
Updated (with thanks to @hannesm, @yomimono and @aantron):

- Only open `Lwt.Infix`, not `Lwt` itself
- Clarify wording in various places
- Make the echo example be about explaining loops
- Improve async/catch explanation
- Move `>|=` explanation higher up because we used it before explaining it
- Moved note about mutexes higher up, just after event loop
- Don't mention that `OS` provides `Main`. We didn't use it, but we do define a different module called `Main`, for maximum confusion.
- Add hyper-links to example sources

talex5:
@yomimono I've moved the event loop stuff much closer to the start than in the original, but still after join and choose. I think it's good to get going with the first challenge before providing too much information at once. I've also moved the "mutex" section up with the event loop - they seem closely related.

yomimono:
I think this looks great.  Thanks again, @talex5!

Issue #183, opened 2014-07-15
"API" link goes to outdated docs


yomimono:
[The "API" link](http://mirage.github.io/) has docs for old versions of packages (e.g., `mirage-tcpip-unix` and `mirage-tcpip-xen` instead of `mirage-tcpip`).

amirmc:
Does running `opam search -s org:mirage` not give us what we need?

PR #461, opened 2016-05-24
Remove API link from navbar.

yomimono:
The API link in the navbar goes to [a very old Github Pages automatically generated repository](https://github.com/mirage/mirage.github.io).  Until we can figure out a nice way to automatically generate this content, remove the link so users aren't directed to outdated documentation.

### DIFF
--- a/src/pages.ml
+++ b/src/pages.ml
@@ -55,7 +55,6 @@ module Global = struct
     tag "ul" ~cls:"left" (list [
       tag "li" (a ~href:(uri "/blog/") (string "Blog"));
       tag "li" (a ~href:(uri "/docs/") (string "Docs"));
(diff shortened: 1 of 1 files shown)


yomimono:
See #183 and https://github.com/mirage/mirage.github.io/issues/1 .

hannesm:
while the index might be outdated, each repository of the mirage organization providing a gh-pages branch will have that online under mirage.github.io/packagename... and e.g. the [functoria documentation](https://mirage.github.io/functoria/) is great over there... shouldn't the fix be a better/working way to generate the index page?

samoht:
It's better to not direct people to outdated contents, so merging that. Individual packages can still still to their own documentation pages. Would be better to have a way to generate consistent documentation across all MirageOS packages (with cross-references). We are almost there so I'ld be happy to add back that link when we have that.

PR #468, opened 2016-07-13
Refresh the installation instructions to something recent

### LINKED ISSUE #467: mirage-www install instructions out of date
- [x]  min is ocaml 4.02.3 not 4.02.1
https://lists.xenproject.org/archives/html/mirageos-devel/2016-07/msg00013.html

### Requirements

* "MirageOS has been tested on ArchLinux, Debian Wheezy, Ubuntu Precise/Lucid/Raring/Saucy, CentOS 6.4 and MacOS X 10.8 and 10.9. To compile the Xen backend, you must have a 64-bit Linux host. 32-bit is not supported at this time."

- [x] Is this out of date information?

* The following points are specific to individual setups and can be separated/sectioned to make for easy, targeted reading:

- [x] "If you're using MacOS X 10.9 or older, you will also need the tuntap kernel module if you want to use the MirageOS network stack from userspace."

- [x] "If you're using Ubuntu/Debian, we recommend you also install the essential build tools (GNU make, etc) and GNU M4 as well as the OCaml toolchain including camlp4:

 apt-get install build-essential m4 ocaml ocaml-native-compilers camlp4-extra"

- [ ] "On CentOS 6.4, install the system compiler and libraries via:

 sudo yum groupinstall "Development Tools" "Development Libraries""

- [ ] "On Arch Linux, install the system compiler and libraries via:

 sudo pacman -S ocaml camlp4 ocaml-findlib
 yaourt -S opam"

### Using OPAM

* "Note that you require OPAM 1.2 or greater to use with MirageOS. Some distribution packages provide earlier versions and must be updated; check with

 $ opam --version ## response should be at least 1.2.0, viz.
 1.2.0"

- [x] Is this out of date? Could be under a subheading of versions? Or have it under requirements?

* "Next, make sure you have at least OCaml 4.01.0 or higher as your active compiler. This is generally the case on MacOS X, though Debian only has it in the unstable distribution at present. But don't worry: if your compiler is out of date, just run opam switch to have it locally install the right version for you.

- [x] Same as above point - can this be in a separate area - or under requirements?

avsm:
Joint work with @GemmaG at hackathon, partially fixes #467
Also bumps minumum OCaml version after report from
John P McDermott on
https://lists.xenproject.org/archives/html/mirageos-devel/2016-07/msg00013.html

### DIFF
--- a/tmpl/wiki/install.md
+++ b/tmpl/wiki/install.md
@@ -21,8 +21,8 @@ If you are using Homebrew, run
 
 This has the latest packages required in the base distribution, so just run:
 
-    apt-get update
-    apt-get install opam
+    sudo apt-get update
+    sudo apt-get install opam
     opam init
     opam install mirage
 
@@ -30,9 +30,9 @@ This has the latest packages required in the base distribution, so just run:
 
 The version of OPAM in older Ubuntus is not high enough to run Mirage (which requires OPAM 1.2.2 or higher), so you will need to add a custom PPA for the latest packages:
 
-    add-apt-repository ppa:avsm/ppa
-    apt-get update
-    apt-get install ocaml ocaml-native-compilers camlp4-extra opam
+    sudo add-apt-repository ppa:avsm/ppa
+    sudo apt-get update
+    sudo apt-get install ocaml ocaml-native-compilers camlp4-extra opam
     opam init
     opam install mirage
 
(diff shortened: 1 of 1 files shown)


tmpl/wiki/install.md
```diff
@@ -1,47 +1,86 @@
...
-After installation, `opam update -u` refreshes the package list and recompiles packages to the latest versions.
+Also note that the `mirage` configuration and deployment tool relies on the `xl` Xen toolstack to run Xen virtual machines.  Older Ubuntus may use the `xm` toolstack, so you will need to change it.
+
+### Debian 
+
+#### Debian Stable (Jessie)
+
+Debian Jessie only packages OPAM 1.2.0, but Mirage needs OPAM 1.2.2 or higher.  You can use [0install](http://0install.net) to get the right version of OPAM:
+
+    apt-get update
```
talex5: Might be worth putting `sudo` before the `apt-get` commands (but not the others).

avsm: done

PR #475, opened 2016-07-19
Clarify section on deprecated functions in Functoria release.

yomimono:
The previous wording implied that `--unix` and `--xen` were deprecated,
but only to those familiar with Functoria's innards.  Add a line making
this more explicit for the `mirage` user.

### DIFF
--- a/tmpl/wiki/breaking-changes.md
+++ b/tmpl/wiki/breaking-changes.md
@@ -9,7 +9,7 @@ an argument).
 
 #### Deprecation: Misc functions
 
-`get_mode` is deprecated. You should use keys instead. And in particular `Key.target` and `Key.is_xen`.
+`get_mode` is deprecated. You should use keys instead. And in particular `Key.target` and `Key.is_xen`.  Users running `mirage configure` should use `-t xen` and `-t unix` instead of `--xen` and `--unix`.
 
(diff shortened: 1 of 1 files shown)


amirmc:
Thank you!

Drup:
Hum, that's actually not correct. `--unix` and `-t unix` are exactly equivalent, there is no difference between them. The whole `get_mode` things come later.

Do we really want to deprecate `--unix` ? It's not hack or anything, Functoria_key.Alias are made specially for that purpose. 

If we do deprecate it, then should we do the same for Alias ?

yomimono:
Oh hm, that's awkward :P thanks for chiming in.  It's clear from your comment that I was confused about `Alias`, which I thought was part of the `base_context` ball of stuff to remove.  If there's nothing harmful about it, I see no need to deprecate it.

samoht:
The alias bit is a bit of a hack as well, so maybe we should deprecate it too? (no strong opinion here)

Drup:
@samoht true, and Alias hasn't seen a lot of use. It could be used for other things though, such as a `--release` CLI flag that would automatically set a bunch of other options. I'm not sure if that's actually useful, I only remember you requested it. :p

The API for aliases is currently very constrained (only flags and only configure-time), mostly due to the fact that I couldn't figure out the semantics for some edge cases.

amirmc:
(For the sake of completeness) There was a brief discussion on the mailing list and the consensus is to deprecate.  See the tail end of the thread — at https://lists.xenproject.org/archives/html/mirageos-devel/2016-07/msg00062.html

yomimono:
I'm going to close out this PR since it's inaccurate.

PR #477, opened 2016-07-19
Replace `--unix` with `-t unix`, and `--xen` with `-t xen`.

yomimono:
Meeting notes and blog entries have been left as is.

### DIFF
--- a/container-run.sh
+++ b/container-run.sh
@@ -10,6 +10,6 @@ export OPAMYES=1
 eval `opam config env`
 opam depext -ui mirage
 cd src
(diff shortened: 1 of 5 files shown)


amirmc:
Would it make sense to update the scripts in this PR too?  e.g https://github.com/mirage/mirage-www/blob/master/container-run.sh?

yomimono:
Yes, I've amended the commit to include that.  We need a bit more discussion after https://github.com/mirage/mirage-www/pull/475#issuecomment-233802501 though.

Issue #476, opened 2016-07-19
Replace uses of '--unix' and '--xen'


amirmc:
Instances of `--unix` and `--xen` should be replaced with `-t unix` and `-t xen` respectively. Related to #475.

Some obvious pages I found (there may be others):
- https://mirage.io/wiki/hello-world
- https://mirage.io/wiki/mirage-www

yomimono:
See #477.  There are others, but I thought it best to leave them as is, since they're historical artifacts:

```
~/mirage-www$ grep -r -E -- "(--xen|--unix)" *
tmpl/wiki/weekly/2015-09-30.md:shorthand flags we have for `--xen` and `--unix` as part of the configure step.
tmpl/blog/intro-tcpip.md:your Mirage project for Unix with `mirage configure --unix`.  This is the mode
tmpl/blog/intro-tcpip.md:`direct` this works with both `mirage configure --xen` and `mirage configure --unix`
tmpl/blog/introducing-vchan.md:    mirari configure --xen --no-install
tmpl/blog/introducing-vchan.md:    mirari build --xen
tmpl/blog/introducing-vchan.md:    sudo mirari run --xen
tmpl/blog/update-on-vchan.md:$ mirage configure --xen
```

amirmc:
Yup, I agree. Thank you :)

amirmc:
There are also references in `mirage-skeleton` that should be updated.  I'll submit a PR there too (unless that's already been done).

```
$ grep -r -E -- "(--xen|--unix)" *
README.md:    $ mirage configure [config.ml] [--unix|--xen]
README.md:The boot target is selected via `--unix` or `--xen`. The default is selected
README.md:This will either execute the native binary created (if on `--unix`) or create
README.md:a default `.xl` configuration file (if on `--xen`). In the latter case you
block/README.md:mirage configure --xen
dns/README.md:$ mirage configure --unix --net=socket
static_website/README.md:$ mirage configure --unix --net=socket
static_website/README.md:$ mirage configure --xen --dhcp=true
```

PR #478, opened 2016-07-19
Replace --(some target) with -t (some target)

amirmc:
The former is deprecated and the latter
already works with the recent releases.

Addresses #476


amirmc:
The changes in #477 cover more files so I'm closing this one :)

PR #486, opened 2016-09-01
use pclock, rather than clock, needed for mirage 3.0

yomimono:


yomimono:
@djs55, this PR causes the `xen` backend tests to fail on reinstalling `c3` (see [here](https://travis-ci.org/mirage/mirage-www/jobs/156920749#L8740)).  I don't understand that bit of the site very well; before I dig in, do you have some sense of what might need to be fixed?

djs55:
This is very odd, I've never seen it before:
```
mirage =>
  cd /home/travis/build/mirage/mirage-www/src && ocamlbuild -use-ocamlfind -tags annot,bin_annot -pkg mirage config.cmxs
Fatal error: exception (Invalid_argument "[ocamlgraph] in_degree")
make: *** [configure] Error 2
'OPAMYES=1 export OPAMYES; set -ex; MODE=$MIRAGE_BACKEND make configure' exited 2. Terminating with 2
```

yomimono:
OK, thanks for taking a look; I'll see whether I can figure it out.

yomimono:
I can reproduce this locally when supplying `--tls false` to `mirage configure -f src/config.ml`.  `--tls true` doesn't trigger the Ocamlgraph exception.

yomimono:
Looking at the `mirage describe --dot` output when `tls` is unspecified, it seems that the default falseness of `tls` is not propagated and there are still nodes in the graph that expect to be supplied with the value of `tls`, whether true or false.  Notably, there are child nodes under `secrets-kv_ro` which can only be reached from `Dispatch_tls`, which we'd expect not to see when `tls` is false.  Orphaning these nodes would violate the invariants stated in `Functoria_graph`, which would lead to the kind of problem we'd expect to generate an exception like `Invalid_argument "[ocamlgraph] in_degree"`.

It seems that providing `mirage configure -t xen --tls false --secrets-kv_ro crunch` (or some other option for --secrets-kv_ro) results in successful completion of `mirage configure`, so perhaps we're on the right track here?

yomimono:
`ocaml-fat` had constraints preventing installation of cstruct >= 2.0.0.  https://github.com/mirage/ocaml-fat/pull/44 should fix this, and hopefully allow tests to pass with the `mirage-dev` repo.

PR #489, opened 2016-09-07
update solo5 url

djwillia:
I received an email by someone who had found the blog post introducing Solo5 and was unable to follow links to the repo, due to its move.  @amirmc is a PR like this the right way to fix it?


yomimono:
I'm no @amirmc but this looks good to me :)

PR #497, opened 2016-10-29
make: make `clean` more robust

mor1:


Makefile
```diff
@@ -46,6 +46,6 @@ build:
 	cd src && make build
 
 clean:
-	cd src && make clean
+	[ -r src/Makefile ] && ( cd src && make clean ) || true
```
hannesm: can't you just `-cd src && make clean` (see https://www.gnu.org/software/make/manual/make.html#Errors)

mor1: Oh, probably-- didn't know that. Is it a GNUism or a standard Make thing?
(Do we care? :)

yallop: or even

```shell
	-make -C src clean
```

mor1: ack; updated again...

PR #500, opened 2016-11-17
Add blog post about Irmin 0.12 from @samoht's announcement

avsm:
…s-devel

### DIFF
--- a/tmpl/blog/irmin-0.12.md
+++ b/tmpl/blog/irmin-0.12.md
@@ -6,7 +6,7 @@ Previously, an Irmin application that wanted to use watches would setup file-sys
   let () = Irmin_unix.install_dir_polling_listener 1.0
 ```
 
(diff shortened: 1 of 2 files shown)


amirmc:
Suggestion: add a final paragraph that says...
```
If you're using Irmin please do let us know how you're getting on via the
[mailing list](https://lists.xenproject.org/cgi-bin/mailman/listinfo/mirageos-devel)
and report any bugs on the [issue tracker](https://github.com/mirage/irmin/issues).
```

avsm:
added, thanks!

tmpl/blog/irmin-0.12.md
```diff
@@ -0,0 +1,21 @@
+Development of the [Irmin](https://github.com/mirage/irmin) Git-like data store continues (see [here](/blog/introducing-irmin) for an introduction). We are releasing [Irmin 0.12.0](https://github.com/mirage/irmin/releases/tag/0.12.0) which brings support for native file-system watchers to greatly improve the performance of watches on the datastore.
+
+Previously, an Irmin application that wanted to use watches would setup file-system scanning/polling by doing:
+
+```
+  let () = Irmin_unix.install_dir_polling_listener 1.0
+```
+
+which would scan the `.git/refs` directory every second. This worked in practise but was unpredictably latent (if unlucky you might wait for a full second for the watch callbacks to trigger), and disk/CPU intensive as we were scanning the full storage directory every second to detect file changes.  In the cases where the store had 1000s of tags, this could easily saturate the CPU. And in case you wonder, there are increasing number of applications (such as [DataKit](https://github.com/docker/datakit)) that do create thousands of tags regularly, and [Canopy](https://github.com/engil/Canopy) that need low latency for interactive development.
```
amirmc: practise -> practice
you wonder, there are increasing -> you were wondering, there are an increasing
[Canopy](https://github.com/engil/Canopy) that need -> [Canopy](https://github.com/engil/Canopy), that needs

PR #505, opened 2016-12-05
update config for exciting mirageos3 developments

yomimono:


src/config.ml
```diff
@@ -77,16 +77,15 @@ let tmplfs = generic_kv_ro ~key:fs_key "../tmpl"
...
 
 let http =
   foreign ~keys "Dispatch.Make"
     (http @-> kv_ro @-> kv_ro @-> pclock @-> job)
 
 let https =
-  let libraries = [ "tls"; "tls.mirage"; "mirage-http" ] in
-  let packages = ["tls"; "tls"; "mirage-http"] in
-  foreign ~libraries ~packages  ~keys "Dispatch_tls.Make"
+  let packages = [package "tls"; package "tls"; package "mirage-http"] in
```
Drup: tls twice ?

yomimono:
I'm puzzled by the failures at https://travis-ci.org/mirage/mirage-www/jobs/181546898 and https://travis-ci.org/mirage/mirage-www/jobs/181546899 .  The build stalls in the `cd src && make depend` phase, but I can't reproduce this locally.

yomimono:
I think this may be an unintended consequence of the pinning workflow -- we're setting keys which are added to the filename of the generated opam file which have `.` in them, which might be interpreted as version numbers -- see `opam pin add --no-action --yes mirage-unikernel-mirage.io-xen .`

yomimono:
PR for the above at https://github.com/mirage/functoria/pull/89 .  Pinning to the fixed branch solves this locally.

yomimono:
The TLS deploy build now fails because `mirage` no longer recognizes `../tls` as a valid argument to `generic_kv_ro` when the `kv_ro` that will be chosen is `fat` (or at least, not `crunch`).

yomimono:
Patch for `invalid_segment` at https://github.com/mirage/mirage/pull/723 but we're still not done; we get a main.ml that doesn't generate a `module Block = ` but tries to reference one.

yomimono:
Further fixes added in mirage/mirage#723 ; this now at least builds locally.  We'll see what Travis thinks once those changes are merged.

yomimono:
At long last our hearts are glad again.

PR #506, opened 2016-12-14
update for fs errors; descend into src/ to run mirage configure

yomimono:
Needed since the merges of https://github.com/mirage/mirage/pull/705 and https://github.com/mirage/functoria/pull/91 .


yomimono:
Test failures should be fixed by https://github.com/mirage/ocaml-fat/pull/57 .

PR #510, opened 2017-01-12
Update to the great `V1` rename

mor1:


yomimono:
At least https://github.com/mirage/mirage/pull/774 is needed, possibly more...

PR #515, opened 2017-02-22
Refresh installation instructions

mato:
- Remove Xen-specific references/replace with "standalone unikernel"
where appropriate
- Remove list of "Linux distributions we test on", replace with generic
"many modern Linux distributions"
- Add backend-specific requirements for Solo5 backends
- Add stub with FreeBSD instructions
- Remove paragraph about upgrading from beta which seems out of date,
not sure if it needs to be here for new users

@yomimono / @avsm Please review, this is a minimal refresh of the installation instructions.

Are we still recommending `0install` as "the way to get opam 1.2.2 on Debian stable"? (/cc @talex5)


yomimono:
Thanks for these updates, @mato. I'm trying to figure out what's suddenly gone wrong with `--net=socket` and may kick this PR for testing a few times, but the text looks good and IMO it's ready to merge.

yomimono:
I was able to reproduce and fix the (unrelated, of course) build error (see https://github.com/mirage/mirage-tcpip/pull/293) , so I think this is good to go.

PR #521, opened 2017-02-23
add API link to navbar: http://docs.mirage.io

yomimono:


yomimono:
Reverts #461 ; we now have nice autogenerated documentation for linking to.

PR #522, opened 2017-02-23
Corrections to hello-world tutorial for ukvm

mato:
- Move the note about examples using the Linux-specific ukvm backend to
a more prominent position as it was buried in the text. Also mention
FreeBSD and GCE.
- solo5 -> Solo5 in a few places
- ukvm-bin is a monitor, not a "program that passes runtime information"
(should maybe link to an explanation but this will do for now)
- Clarify wording of how ukvm-bin passes keys to the unikernel

### DIFF
--- a/tmpl/wiki/hello-world.md
+++ b/tmpl/wiki/hello-world.md
@@ -334,7 +334,9 @@ $ ./hello
 
 #### Building for Another Backend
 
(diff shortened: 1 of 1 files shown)


yomimono:
Thanks very much for the edits, particularly in regard to `ukvm-bin`, which I was at a loss to describe both succinctly (in the context of the examples) and correctly.

PR #552, opened 2017-05-16
Fix tutorial/hello/config.ml module explanation

bramford:
The tutorial/hello/config.ml explanation references the `Main` module several times but only the `Hello` module exists in unikernel.ml


yomimono:
Thanks very much for this fix, @bramford !

PR #553, opened 2017-05-22
Add notes on jbuilder and topkg

djs55:
Following my recent experiments I thought it would be a good idea to write down the various "best practices" or "tips and tricks" for using jbuilder and topkg.

Signed-off-by: David Scott <dave@recoil.org>

### DIFF
--- a/src/data.ml
+++ b/src/data.ml
@@ -768,7 +768,12 @@ module Wiki = struct
       body       = File "breaking-changes.md";
       permalink  = "breaking-changes";
     };
-
+    { updated    = date (2017, 05, 22, 18, 00);
+      author     = dave;
+      subject    = "Building and packaging with jbuilder and topkg";
+      body       = File "packaging.md";
+      permalink  = "packaging";
+    };
   ]
 end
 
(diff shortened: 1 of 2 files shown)


djs55:
The other entries in here look quite old -- perhaps I should be using canopy instead?

yomimono:
This is really great!  Thanks for writing it up. :)

I don't have a strong opinion on where it should live, but there is [a "help needed" project over on Canopy about porting packages/repositories to topkg](http://canopy.mirage.io/Projects/topkg).  Links in both directions would be a good idea IMO.

yomimono:
It looks like [mirage-block-unix 2.8.2](https://github.com/mirage/mirage-block-unix/releases/tag/v2.8.2) should fix the test failure here - thanks @avsm !

avsm:
sorry, that was my bad to start with, so fixing it shortly :)

avsm:
fixed in https://github.com/ocaml/opam-repository/pull/9257 and merged

samoht:
Thanks for your notes! I have a couple of remarks, I will create a new PR to complete your notes with my notes :-)

djs55:
I look forward to reading your improvements :)

PR #561, opened 2017-07-05
rework travis, make etc to use dommage to build

mor1:
also tidies things up to Mirage v3

not for merge yet-- i still need to debug the deployment paths.


mor1:
Ok-- I believe that this should go green now. The final steps in deployment can't really be tested until Travis runs on not-a-PR but the scripts in question seem to work for my site.

@yomimono I think this is good to merge, but you probably should check I didn't miss anything dumb first :)

yomimono:
I'm not familiar enough with dommage to review this properly just by looking at it, and I won't be able to try running it in a reasonable timeframe.

.travis.yml
```diff
@@ -1,23 +1,37 @@
...
+  - docker
 language: c
-install: wget https://raw.githubusercontent.com/avsm/ocaml-travisci-skeleton/mirage3/.travis-mirage.sh
-script: bash -ex .travis-mirage.sh
+cache: false
+
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install -y realpath # needed on ubuntu 14.04.5 LTS / trusty
+  - wget https://raw.githubusercontent.com/mor1/dommage/master/dommage
```
djs55: Perhaps we should pin a commit (or release tag) here rather than tracking master?

Makefile
```diff
@@ -16,36 +16,45 @@
...
-	cd src && make build
+test:
+	$(MIRAGE) run sudo “./www
 
-clean:
-	-make -C src clean
-	cd stats && make clean
-	$(RM) log src/mir-www src/*.img src/make-fat*.sh
+publish:
+	$(MIRAGE) publish mor1/mirage-www
```
djs55: I'm a bit suspicious about the reference to `mor1/mirage-www` here.

.travis.yml
djs55: Is this ... (sharp intake of breath) not an OCaml program?

.travis.yml
```diff
@@ -1,23 +1,37 @@
...
+        --host openmirage.org --redirect https://mirage.io"
+      DEPLOY=1 XENIMG="openmirageorg" MIRDIR="src" UPDATE_GCC_BINUTILS=1
+    - FLAGS="-vv --net direct --dhcp false --ipv4 46.43.42.136/28
+        --ipv4-gateway 46.43.42.129 --tls true --secrets-kv_ro fat
+        --host mirage.io"
+      DEPLOY=1 XENIMG="mirageio" MIRDIR="src" UPDATE_GCC_BINUTILS=1
+
   global:
-  - FORK_USER=avsm
-  - FORK_BRANCH=mirage3
```
djs55: Just checking -- was there any magic still left in the avsm branch?

avsm:
I'd like to understand what problems this caching layer is solving.  We seem to have an issue with rebuilding the website from scratch on every push, but we could also fix it by making mirage-www output static HTML and then just using a basic https unikernel to serve it.  Would it be possible to start a discussion about this so it's seen more widely @mor1?

.travis.yml
djs55: Is the `dommage` tool Mirage-specific? Should it be moved to the Mirage org? Possibly `ocaml/ocaml-travisci-skeleton` was never a great place for Mirage-specific stuff either.

mor1: It's a shell script, the universal language for running things and manipulating files. So sue me ... ;)

(In all seriousness: avoiding the need to get OPAM/OCaml correctly installed on the host Ubuntu image, and instead being able to contain all the OCaml-ness inside the pre-built and reused container seemed to simplify things to me.)

Makefile
mor1: True; ideally that would be under the `mirage` org on Hub.
Happy to start pushing images there instead if the overall approach seems sensible.

mor1:
@yomimono understood, I should've tagged more folk than just you, sorry!
@avsm you mean a discussion on `mirageos-devel` list? Sure, can do.

mor1:
@djs55 Yes, I should write up a blog post about `dommage` anyway. It's sort-of mirage specific currently, though it needn't be I guess. The rationale for writing it was that I was having odd OPAM package install inconsistencies when Travis was trying to rebuild my website for deployment (resulting in the rebuild failing), when compared with what I was getting on my laptop, and I never managed to repro locally. So I thought that containerising the build environment so that it appears to be Linux in both cases might help, and then realised that it would also mean I could cache the build environment and push it to Hub for Travis to pick up, meaning Travis didn't need to recreate it each time.

@avsm the problem is not so much rebuilding the site itself (`mirage/mirage-www`) every time, it's the reinstallation (by building from source) of the ~75 or so OPAM packages that the site depends on. That's what pushed the Travis build time up from ~4min to ~20min. If the entire site was refactored so that we use Jekyll or something to generate and then commit the static content, and simply serve it out of a basic unikernel webserver, that would work too. But this way seemed easier and still makes `mirage-www` a reasonably "large" sample unikernel.

.travis.yml
mor1: PS. Yes absolutely I should make this a proper release.

talex5:
It looks like this PR has stalled a bit. I have an alternative deployment system using OCurrent here:

https://github.com/ocurrent/ocurrent-deployer/blob/d007d0bb08320d4245effdbd2a2ee43b8a1cb442/src/pipeline.ml#L67

It monitors the `live` branch of this repository, builds it when it changes, and deploys it (currently to http://147.75.34.226/). The unikernel image deployed is whatever the `Dockerfile` builds with `--build-arg TARGET=hvt`, so it should be easy to reproduce locally.

To be useful, this needs a bit of updating to deal with https certificates (at the moment, it's http-only). We could add a pipeline stage to run Let's Encrypt as necessary, or perhaps embed our own Mirage client inside the unikernel.

The build should be reasonably fast when making multiple changes because of Docker's build caching (dependencies are installed in a separate layer that can be reused when the content changes).

avsm:
Superseded by the ocurrent workflow in #673

PR #572, opened 2017-12-02
Fix for Cohttp interface changes

mato:
Thanks @engil for the help


mk270:
The Travis stuff probably wants to be its own PR

yomimono:
These changes are included in the merged #574 .

PR #578, opened 2017-12-17
further deploy script massaging

yomimono:


yomimono:
Merging this myself so I can keep an eye on it; there may be follow-up PRs.  Sorry for the noise, but I have no better way of debugging this.

PR #599, opened 2018-01-26
blog post on esp32 from Sadiq Jaffer

avsm:
cc @sadiqj


hannesm:
oh wow that's great to hear! esp32 is a great platform (cheap, wifi included) :D

avsm:
Yeah, I'm ordering in a few -- will physically mail them around when the hardware arrives :-)

PR #608, opened 2018-05-04
Update info for Amir

amirmc:
Removed affiliation and also dropped the title.
That particular function is spread across a number
of folks, so it's been outdated for some time.


yomimono:
Thanks, @amirmc !

PR #610, opened 2018-05-16
Update to cow 2.3.0 to fix the atom feed

samoht:


mor1:
LGTM! :) 

But until someone updates my public key, I can't login to the deployment box to fix things in case something breaks during deployment so I won't merge...

samoht:
Hum the atom feed is still broken .... I guess this needs a bit more investigation after all :-/

PR #614, opened 2018-08-14
Fix Irmin examples

zshipko:
This ranks pretty high when searching for Irmin, it should definitely be up-to-date!


samoht:
Thanks!

PR #615, opened 2018-09-17
Update ARM64 wiki page to reflect issues mentioned in issue 919

derpeter:
As suggested in https://github.com/mirage/mirage/issues/919#issuecoment-421879185 i mentioned the gcc version needed to build the image

and also mentioned that Arch Linux could be used instead of building you own debian image.


hannesm:
thanks, merging

PR #626, opened 2019-03-10
Fix some broken links

rymdhund:
Here is a fix for all the broken links internal to mirage.io that I could find. And some of the external ones. Some relative paths were working on the wiki, but were broken on the the atom feed. 

I moved the path for the static documents to `/documents/` since `/docs/` is overridden.


hannesm:
thanks!

PR #627, opened 2019-03-15
reyn.ir -> reynir.dk

reynir:
The domain reyn.ir has become unavailable due to sillyness around US
sanctions against Iran and a lost password.


hannesm:
thanks!

PR #669, opened 2020-03-19
Fix deployment

talex5:
- Travis was failing due to https://github.com/ocaml/ocaml-ci-scripts/pull/321.
- Docker was failing due to the 4.09.1 release and https://github.com/mirage/mirage/pull/1086.


samoht:
Thanks!

PR #682, opened 2020-07-09
Changing 404ing link to working docs link

P-Jon:
Currently the link to documentation is broken - I'm unsure whether this is because it was forgotten to be changed to the docs link I could find or a different problem.


talex5:
It used to point here: https://web.archive.org/web/20200122150121/https://docs.mirage.io/

Looks like https://github.com/mirage/docs/issues/4 is the cause (missing `index.html`).

hannesm:
thanks for your report, @dinosaure fixed the https://docs.mirage.io in the meantime -- thus closing this issue. if it still does not work for you, please let us know.

PR #687, opened 2020-10-26
Treat letsencrypt-related options as required only when TLS is enabled

mato:
Treat `--dns-key`, `--dns-server`, `--dns-port` and `--key-seed` as required only when TLS has been enabled at configure time (with `--tls=true`).

This allows for easily running the website unikernel locally for preview purposes.

I'm not 100% sure that the current failure mode is the best way to go about it; if `--tls` is configured but the options are not given, then you get this:

```
2020-10-26 12:06:57 -00:00: INF [netif] Plugging into service with mac 16:bf:19:7c:1b:40 mtu 1500
2020-10-26 12:06:57 -00:00: INF [ethernet] Connected Ethernet interface 16:bf:19:7c:1b:40
2020-10-26 12:06:57 -00:00: INF [ARP] Sending gratuitous ARP for 10.0.0.2 (16:bf:19:7c:1b:40)
2020-10-26 12:06:57 -00:00: INF [udp] UDP interface connected on 10.0.0.2
2020-10-26 12:06:57 -00:00: INF [tcpip-stack-direct] stack assembled: mac=16:bf:19:7c:1b:40,ip=10.0.0.2
Fatal error: exception (Invalid_argument "--tls is enabled but required key missing, see --help")
Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
Called from Dispatch_tls.Make.required in file "dispatch_tls.ml" (inlined), line 72, characters 16-83
Called from Dispatch_tls.Make.tls_init in file "dispatch_tls.ml", line 77, characters 6-41
Called from Dispatch_tls.Make.start in file "dispatch_tls.ml", line 92, characters 4-48
Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
Called from Main in file "main.ml", line 220, characters 2-240
Solo5: solo5_exit(2) called
``` 

What's a less verbose and more obvious way to exit in this case?

Also, can I get the actual missing key name into the exception error message somehow? If so, how?

### DIFF
--- a/src/dispatch_tls.ml
+++ b/src/dispatch_tls.ml
@@ -67,7 +67,7 @@ module Make
     | _ -> ()
 
   let required key_name key =
-    match (key ()) with
+    match key () with
       | Some x -> x
       | None -> Log.err (fun m -> m "TLS is enabled but required key %s missing" key_name); exit 64
 
--- a/src/dispatch_tls.ml
+++ b/src/dispatch_tls.ml
@@ -66,10 +66,15 @@ module Make
       end
     | _ -> ()
 
+  let required key_name key =
+    match (key ()) with
+      | Some x -> x
+      | None -> Log.err (fun m -> m "TLS is enabled but required key %s missing" key_name); exit 64
+
   let tls_init stack hostname additional_hostnames =
-    C.retrieve_certificate stack ~dns_key:(Key_gen.dns_key ())
-      ~hostname ~additional_hostnames ~key_seed:(Key_gen.key_seed ())
-      (Key_gen.dns_server ()) (Key_gen.dns_port ()) >>= function
+    C.retrieve_certificate stack ~dns_key:(required "--dns-key" Key_gen.dns_key)
+      ~hostname ~additional_hostnames ~key_seed:(required "--key-seed" Key_gen.key_seed)
+      (required "--dns-server" Key_gen.dns_server) (required "--dns-port" Key_gen.dns_port) >>= function
     | Error (`Msg m) -> Lwt.fail_with m
     | Ok certificates ->
       restart_before_expire certificates;
--- a/src/config.ml
+++ b/src/config.ml
@@ -26,7 +26,7 @@ open Mirage
 
 let tls_key =
   let doc = Key.Arg.info
-      ~doc:"Enable serving the website over https. Do not forget to put certificates in tls/"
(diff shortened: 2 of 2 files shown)


hannesm:
Thanks for your PR.

> What's a less verbose and more obvious way to exit in this case?

I'd use instead of `invalid_arg` a log message (logging has been configured before the `start`/`main` is executed): `Log.err (fun m -> m "TLS is enabled, missing key %s" key_name); exit 64`. The `key_name` needs to be passed into `required` AFAICT. The `exit 64` will exit the unikernel with exit code 64 (argument error, as done in https://github.com/mirage/functoria/blob/master/runtime/functoria_runtime.ml#L83 -- eventually that should be defined in Mirage_runtime).

> Also, can I get the actual missing key name into the exception error message somehow? If so, how?

The type for an argument key is a Functoria_runtime.Arg.t https://github.com/mirage/functoria/blob/master/runtime/functoria_runtime.mli#L56 -- which does not provide any API to get the key name. In the meantime, I'd use `let required key_name key = ...` and manually pass the key name. In the longer term, we should either allow for expressing "this key is required if the other key is true" or an API to get the name of the key.

mato:
Okay, PTAL at the latest version, seems to work.

src/dispatch_tls.ml
```diff
@@ -66,10 +66,15 @@ module Make
       end
     | _ -> ()
 
+  let required key_name key =
+    match (key ()) with
```
hannesm, line 70:
```suggestion
    match key () with
```

In OCaml there is no need for parens.

hannesm:
LGTM, I commented a minor nit.

samoht:
Could you try instead to move the keys defined in https://github.com/mirage/mirage-www/blob/master/src/config.ml#L92-L93 to https://github.com/mirage/mirage-www/blob/master/src/config.ml#L107 ? That'll make these keys be defined only for the `https` implementation.

mato:
@samoht That is indeed a better solution, thanks! See #688.

PR #688, opened 2020-10-26
Only enable TLS-related keys if TLS is enabled

mato:
This allows for local runs w/o TLS and depending on the letsencypt
parts. It also makes the list of options less confusing; if
`--tls=false` (the default), then TLS-related options are not shown or
accepted.


samoht:
LGTM

PR #694, opened 2021-04-07
Mirage 4 build

TheLortex:
The main changes are:
- Dockerfile: use the mirage 4 development repository `mirage-dev` (`mirage-4` branch) and adapt a bit to mirage 4 changes.
- omd: switch to beta 2.0.0 as the old one doesn't work with dune -- needed to change some `.md` files as they would produce invalid HTML
- move template files into `src/`, it's required because they need to live inside the dune project. 
- adapt to the new cohttp


src/dispatch_tls.ml
```diff
@@ -81,7 +68,7 @@ module Make
     let redirect = Key_gen.redirect () in
     let hostname = Domain_name.(of_string_exn (Key_gen.host ()) |> host_exn) in
     let additional_hostnames =
-      List.map (fun n -> Domain_name.(host_exn (of_string_exn n)))
+      List.map (fun n -> Domain_name.(raw (of_string_exn n)))
```
hannesm, line 71:
```suggestion
      List.map Domain_name.of_string_exn
```

PR #699, opened 2021-06-03
Add myself as contributor.

haesbaert:


hannesm:
thanks!

hannesm:
in retrospect, the list seems to be mostly in alphabetic order of the last name, could you put yourself at the right place? ;)

haesbaert:
sure, I can fix the order too, and swap for first name, because.

PR #701, opened 2021-07-22
Updated landing page text & changed Canopy menu link to Irmin.org

christinerose:
Unable to build with Docker yet, but I will try again later. 
Found the correct MD files, and changed those, but both intro-1.md and intro-2.md had some of the same info. Since I wasn't able to see which fed into the build, I just changed them both with the Irmin info. 

I really took a chance on editing the src/pages.ml to change the broken Canopy link to Irmin. I hope I got it right! If not, then there's something else for me to learn. :)

### DIFF
--- a/src/pages.ml
+++ b/src/pages.ml
@@ -57,7 +57,7 @@ module Global = struct
       tag "li" (a ~href:(uri "/blog/") (string "Blog"));
       tag "li" (a ~href:(uri "/docs/") (string "Docs"));
       tag "li" (a ~href:(uri "http://docs.mirage.io") (string "API"));
-      tag "li" (a ~href:(uri "http://canopy.mirage.io") (string "Canopy"));
+      tag "li" (a ~href:(uri "http://irmin.org") (string "Irmin"));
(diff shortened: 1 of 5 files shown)


hannesm:
Thanks for the PR, some notes:
- why advertise irmin.org prominently at 3 places (including the header)? at least for me, irmin is a nice project, but there are lots of use cases without irmin. I consider e.g. the tcpip stack more central than irmin.
- I'm confused by "or as a pure JavaScript library inside a browser" -- does this work? if so, are there instructions available?

dinosaure:
> I'm confused by "or as a pure JavaScript library inside a browser" -- does this work? if so, are there instructions available?

I agree with this point. At this stage, we only experiment few libraries and see if we can get the JavaScript support but, into our development process or our tools, we don't have anything to strictly say that it's working.

christinerose:
No worries! We don't have to put Irmin in the menu, but the Canopy link was broken. I'd suggest to either remove that link all together or to replace it with something the team thinks works in the menu.

PR #692, opened 2021-02-05
Adapt to mirage 3.11

samoht:
Testing my conduit changes.

PR #702, opened 2021-09-10
pass a stack v4v6 to conduit

yomimono:
`conduit_direct` expects a dual-stack argument these days.


talex5:
CI is probably failing because the opam-repository commit hash in `Dockerfile` needs updating too.

dinosaure:
Thanks @yomimono for your look 👍 it's really pleasant to see that you can reproduce `mirage-www` with MirageOS 4 and thanks for your time on that. I added some minor patches and it's ready to merge (and close #692). I will unlock the 4.12 support with your `cowabloga` patche.

PR #707, opened 2022-01-27
Misc polishing of the new website

tmattio:
This PR addresses some TODOs on the next version of mirage.io.

The commit history should be descriptive and the commits self-contained.

@TheLortex I still had issues installing the opam deps with a fresh opam switch. I think https://github.com/TheLortex/dream/pull/1 should help, but I also removed the yaml pin. Could you add it to the mirage config?

I also had to fix a module name in `bin/main.ml` (`Unix_os` -> `OS`), I'm assuming this is because of a different version of a dependency. We may need some constraints on the dependencies.

Let me know if you spot anything on the new website (https://next.mirage.io/) and I can address it in this PR as well.


TheLortex:
Thanks, I've added some commits:
- to take  https://github.com/TheLortex/dream/pull/1 into account in the pin-depends
- add dependencies for `bin/main.ml` in a new `mirageio-bin` package
- move the `pin-depend` on `yaml` to the mirage config
- add some documentation on how to build the unikernel

TheLortex:
Merging as the unikernel builds !

PR #708, opened 2022-01-28
Implement caching for static files

TheLortex:
Caching is hard, so as a first implementation I've added the following, only for files that are served statically:
- new response header `Cache-Control: max-age=3600`: static files are cached for one hour, once the website gets stabilized we could bump that up to 1 day or more
- new response header `Etag: "xxx"` is a tag that is used so that the client can ask, when the resource has expired, whether the content has actually changed. Currently, the tag is the hash of all the static files. At some point we could make it so that crunch computes the hashes for each file. For now it's a good enough approximation.
- parse request header `If-None-Match: "xxx"` that should return 304 not modified if the Etag matches what the client has in cache.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,2 +1,3 @@
 version = 0.20.1
 profile = default
+ocaml-version = 4.10.0
--- a/dune-project
+++ b/dune-project
@@ -25,6 +25,7 @@
   dune
   dream
   dream-mirage
+  mirage-kv-mem
   mirage-clock-unix
   mirage-unix
   (crunch :build)
--- a/lib/dune
+++ b/lib/dune
@@ -1,9 +1,9 @@
 (library
  (name mirageio)
- (libraries dream-mirage mirage-time tcpip mirageio_template mirageio_data))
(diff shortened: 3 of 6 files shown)


hannesm:
sounds good -- though what you can as well use is the last_modified (crunch emits at module creation time the current timestamp, and each value in the mirage-kv store has this as modification-time). this value can also be used for the last-modified header (and if-modified-since could be interpreted).

hannesm:
FWIW I have some similar code in https://github.com/roburio/unipi/blob/main/unikernel.ml#L20-L61 -- which uses the git implementation (not crunch)

TheLortex:
Thanks for the pointer, because `crunch` is used in _plain_ mode I didn't see that a `last_modified` field was available.

Do you think implementing the two mechanisms (Etag and Last-modified) is useful for now ?

hannesm:
I think either is sufficient. Since you add a call to "sha256sum" that is not available on all systems (e.g. only on some Linux systems, others need specific packages -- on SD it is named sha256), I'd appreciate to remove that and use the last modified timestamp :)

TheLortex:
Now using the last modified timestamp !

PR #714, opened 2022-02-11
Add a /metrics endpoint for monitoring

TheLortex:


TheLortex:
Note: I have extracted some code from https://github.com/mirage/prometheus/blob/master/app/prometheus_app.ml because the `prometheus-app` package is depending on cohttp that we don't need here.

PR #715, opened 2022-02-14
fixed wiki links

RyanGibb:


avsm:
This change looks good, but shouldn't there be a redirect to the old URLs also in place for external links, @samoht @dinosaure ?

Also need to investigate why ocaml-ci is failing on the multicore package.

TheLortex:
Merging this PR, and I will add the `wiki/ => docs/` redirection.

PR #716, opened 2022-02-18
Fix pin-depend on 'yaml'

Julow:
The constraints have changed in opam and we need to specify the pinned version of `yaml`. This requires a patch to mirage: https://github.com/mirage/mirage/pull/1295

Perhaps we need a workaround that doesn't use a patched version of mirage ?

### DIFF
--- a/mirage/config.ml
+++ b/mirage/config.ml
@@ -106,7 +106,7 @@ let packages =
     package ~build:true
       ~pin:
         "git+https://github.com/TheLortex/ocaml-yaml.git#7e1f117645ea10fbec2bd3dbbf0d8f581cce891f"
-      "yaml";
+      ~pin_version:"3.0.0~dev" "yaml";
   ]
 
 let https =


mirage/config.ml
```diff
@@ -106,7 +106,7 @@ let packages =
     package ~build:true
       ~pin:
         "git+https://github.com/TheLortex/ocaml-yaml.git#7e1f117645ea10fbec2bd3dbbf0d8f581cce891f"
-      "yaml";
+      ~pin_version:"3.0.0" "yaml";
```
hannesm, line 109: I'd propose to use `3.0.0.dev` or `3.0.0-dev` or `3.0.0~dev` to indicate for later that this wasn't the bare 3.0.0 version.

Julow: Which one ? :) mirage uses `~` for the beta versions so I did that.

hannesm: Depends on the actual constraints, since they're ordered below and above 3.0.0 ;)

mirage-runtime.4.0 < mirage-runtime.4.0.0~beta < mirage-runtime.4.0.0

avsm:
Is this pin needed at all? I released 3.0.1 of yaml to opam-repo.

Julow:
The pin is still needed due to a constraint in `ppx_deriving_yaml`:
```
               - ppx_deriving_yaml -> (problem)
                   Rejected candidates:
                     ppx_deriving_yaml.0.1.0: Requires yaml < 3.0.1
```

and otherwise, with the stock yaml `3.0.0`, I get this when building:

```
File "mirage/duniverse/ocaml-yaml/types/stubgen/dune", line 11, characters 0-232:
11 | (rule
12 |  (targets ffi_ml_types_stubgen.exe)
13 |  (deps    (:c ./ffi_ml_types_stubgen.c) ../../vendor/libyaml_c_stubs.a)
14 |  (action (run %{cc} %{c} -I../../vendor %{read-lines:../../config/ctypes-cflags} 
15 |   -I %{ocaml_where} -o %{targets})))
ffi_ml_types_stubgen.c:10:10: fatal error: ctypes_cstubs_internals.h: No such file or directory
   10 | #include "ctypes_cstubs_internals.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

avsm:
cc @patricoferris -- could we get a new version of ppx_deriving_yaml released then? I think this is just a minor change due to the rresult/result removals.

TheLortex:
`ocaml-yaml` doesn't work in a monorepo context so I had to perform a patch: https://github.com/TheLortex/ocaml-yaml/commit/7e1f117645ea10fbec2bd3dbbf0d8f581cce891f

avsm:
That fix doesn't look right -- I think this should be fixed in the ctypes port and not in yaml, since that's where the integers inclusion should come from...?

TheLortex:
No it's just a temporary fix as dune 3 is supposed to fix the ctypes issue. So at some point we'll properly switch to using `yaml` with proper ctypes support in dune: https://github.com/avsm/ocaml-yaml/pull/41 but currently the ctypes work in dune 3 doesn't work in monorepos when the C library is vendored

TheLortex:
I see this PR in your dune port of ctypes: https://github.com/avsm/ocaml-ctypes/pull/4 so there's probably a better way to fix the problem in the meantime.

patricoferris:
> cc @patricoferris -- could we get a new version of ppx_deriving_yaml released then?

Yep https://github.com/ocaml/opam-repository/pull/20814

TheLortex:
So https://github.com/avsm/ocaml-ctypes/pull/4 has been merged. 

I have opened  https://github.com/avsm/ocaml-yaml/pull/52 to make ocaml-yaml in monorepos without breaking anything else (especially opam installation).

hannesm:
What is the state of this? Do we need it? It looks like the main branch works without this patch (maybe since some packages have been released to opam-repository in the mean time)?

TheLortex:
Yes, we've got a new version of `ppx_deriving_yaml` on opam-repository so the version doesn't have to be pinned anymore.
In any case thanks @Julow for the patch on the mirage tool, this is still an useful feature.

PR #718, opened 2022-03-06
Fix lexicographical order and tidy things up.

haesbaert:
Spotted a couple of entries out of order, hopefully didn't miss anything.
While here, keep the same style on every entry.


haesbaert:
any love on this ?

dinosaure:
Thanks, some issues remains about our fork of `dream` (/cc @TheLortex) but this PR seems fine.

PR #724, opened 2022-03-26
Use a packaged version of tailwindcss

samoht:


samoht:
The `Pext_decl` issue should be fixed by https://github.com/ocaml/opam-repository/pull/21019

samoht:
And the other one should be fixed by https://github.com/mirage/mirage-www/pull/721

samoht:
Also the Dockerfile entry is still needed because of https://github.com/tmattio/opam-tailwindcss/issues/1

PR #725, opened 2022-03-26
Use latest mirage-dev to fix deployment

samoht:


samoht:
Deploy should be fixed by https://github.com/ocaml/opam-repository/pull/21022

PR #737, opened 2022-03-30
Rephrase weekly calls as in the past

reynir:
It's been quite some years since the last fortnightly call. I rephrased the text above the list of fortnightly calls a little bit so it is more clear it's a thing of the past.


samoht:
Good point! Would be great to restart these calls :-)

PR #746, opened 2022-04-07
Add RSS auto-discovery

samoht:
Following https://dri.es/rss-auto-discovery

### DIFF
--- a/template/layout.eml
+++ b/template/layout.eml
@@ -1,9 +1,15 @@
-type tab = 
+type tab =
   | Home
   | Blog
   | Docs
   | Api
   | Community
+
+let extra_link = function
+  | Blog ->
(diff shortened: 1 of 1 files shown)


template/layout.eml
```diff
@@ -1,10 +1,16 @@
-type tab = 
+type tab =
   | Home
   | Blog
   | Docs
   | Api
   | Community
 
+let extra_link = function
+  | Blog ->
```
TheLortex, line 9: Maybe it'd be useful to put it on all pages so that even people visiting the home page are able to discover the RSS feed ?

I see that https://dri.es is doing it so it looks like standard practice.

reynir: Yes, I agree. You can put multiple rss feed links in the header. Just use an informative title :-)

PR #754, opened 2022-04-20
Fixed code block formatting for the remaining posts

davisuga:
Added missing tags for OCaml code


samoht:
Thanks!

Out of curiosity, what are you using these tags for? As unfortunately our current web engine just skips those...

davisuga:
I'm planning to patch the current web implementation to include syntax highlighting. Also, you can see the highlighting via GitHub or VSCode MD preview

samoht:
Great! Looking forward for your patches :-)

PR #759, opened 2022-05-30
Cross-compilation is working with dune 3

TheLortex:
Re-running the tutorial from scratch, it looks like recent updates of dune 3 have fixed the cross-compilation issues.


hannesm:
Shouldn't we then specify a concrete lower bound for dune (`>= 3.2.0`)? To avoid that someone tries to install and due to some solver issues (or an arcane opam-repostory) installs a dune version that does not work.

dinosaure:
Currently, `mirage` requires `dune.3.0.0` for tests so it seems fine to definitely require `dune.3.0.0` for the mirage tool (and handle correctly the cross-compilation).

TheLortex:
dune 3.0.0 didn't work with cross-architecture compilation, hence the current state of the blog post advising to use dune 2. Now a safe lower bound is 3.2.0.

hannesm:
So maybe a lower bound of 3.2.0 should then be added to the respective releases & opam files?

PR #762, opened 2022-08-22
Update opam-repository commit

TheLortex:
Server has run out of memory. I'm proposing an update of the opam repository commit to integrate memory leak fixes in mirage-tcpip 7.1.2


TheLortex:
I'm also removing the mirage 4.0.0 pin as it's not working anymore. We're now using mirage 4.2.0

PR #767, opened 2022-10-04
Use mirage-monitoring to monitor the unikernel and provide a memprof endpoint

TheLortex:
[This ](https://github.com/roburio/mirage-monitoring) is used to provide metrics reporting to InfluxDB and a TCP endpoint to dump the memprof trace.

The PR adds two configure-time options:
- `--metrics` to enable mirage-monitoring.
- `--separate-networks` to avoid exposing the monitoring network to the internet (as the influxdb sink is not secured)

The deployer has to be updated : https://github.com/ocurrent/ocurrent-deployer/pull/143

### DIFF
--- a/mirage/config.ml
+++ b/mirage/config.ml
@@ -124,9 +124,92 @@ let app =
       (Letsencrypt, https $ default_random);
     ]
 
+let separate_networks =
+  let doc =
+    Key.Arg.info ~doc:"Separate external and internal networks"
+      [ "separate-networks" ]
+  in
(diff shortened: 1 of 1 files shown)


TheLortex:
Failure on Xen because `mirage-monitoring` only works with `mirage-solo5`. As both `mirage-solo5` and `mirage-xen` are pulled it duplicates symbols.

hannesm:
FWIW if there's interest, we can likely make mirage-monitoring working with mirage-xen as well...

Issue #772, opened 2022-11-08
mirage.io/community/ returns an blank 404 page


kit-ty-kate:
* `mirage.io/community` works fine
* `mirage.io/community/` returns a blank page
* `mirage.io/communit` returns a proper 404 page

note that `community/` is the one i got redirected to from duckduckgo

PR #788, opened 2023-01-04
Handle trailing slashes

TheLortex:
Fixes https://github.com/mirage/mirage-www/issues/772

Code from ocaml.org website: https://github.com/ocaml/ocaml.org/blob/main/src/ocamlorg_web/lib/middleware.ml

Made on top of https://github.com/mirage/mirage-www/pull/785

### DIFF
--- a/lib/middleware.ml
+++ b/lib/middleware.ml
@@ -26,4 +26,14 @@ struct
           ~headers:(Dream.all_headers response)
           (Dream.status response)
     | _ -> handler request
+
+  let no_trailing_slash next_handler request =
+    let target = Dream.target request in
+    match target with
+    | "/" -> next_handler request
+    | _ ->
+        let length = String.length target in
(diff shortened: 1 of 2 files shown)


reynir:
This change allows you to craft URLs at mirage.io that redirect to arbitrary URLs, e.g https://mirage.io///github.com/mirage/mirage-www/pull/788/. I'm not sure what's the best way to fix that. It can also lead to too many redirect: https://mirage.io/blog///////////////////////////////////////////////////////////////

I am also a bit wary of removing trailing slashes in general as it breaks relative URLs in pages, but that may not be a problem in mirage-www(?)

TheLortex:
Thanks for the review, and sorry for merging too fast.

I think all links in the website are absolute, so it's not breaking to remove trailing slashes. 
However the redirection needs to be smarter indeed.

Issue #779, opened 2022-12-06
new website: security advisories are duplicated


hannesm:
there are now several versions available:
- tmpl/advisories
- data/blog/MSA*md

is this really needed? what is the story of keeping them in sync? this is pretty inconvenient for adding new data (also for editing..).

TheLortex:
One is the PGP-signed message (in `tmpl/advisories`), the other is processed by the website's CMS. 
It should be possible to only keep the PGP-signed messages and build blog articles for them.

PR #791, opened 2023-01-04
Deduplicate security advisories by generating blog articles from the PGP messages

TheLortex:
A solution for https://github.com/mirage/mirage-www/issues/779

Source of truth is PGP signed messages. The site generator reads them and automatically generates blog post articles according to the content. In particular is looks for the `Announced` field to set the blog post date, and the `Affects` field for blog post subject. 

For consistency I have moved the advisories under the `data/security` folder. I have also edited the content of the advisories so that they have a consistent style, that is easier to parse. This means that the PGP signatures are not correct anymore, @hannesm you will have to update the signatures.


hannesm:
Dear @TheLortex, thanks for this PR. But since PGP signatures carry a timestamp, I'm hesitant to re-sign all advisories without adding any information inside of them.

In addition, some of your formatting changes are done by the sign operation (such as adding trailing whitespaces, but also leading dash handling is special). Did you test that removing the signature and signing the raw data will lead to something your ad-hoc parser is able to use?

I would also refrain from moving the advisories to another random place once again - I don't see any benefits in doing that, but please elaborate why this is needed?

Issue #792, opened 2023-01-12
struggle with compiling it locally


hannesm:
Hello,

thanks for the new design! :) Unfortunately, it looks like the new architecture and design requires the opam package "tailwindcss" (https://github.com/tmattio/opam-tailwindcss), which according to the opam file is distributed as binary for GNU/Linux and macOS. This opam package is also pinned to a specific commit of the above repo, so I wonder whether anyone keeps track of updates.

I'm using FreeBSD, and am curious what tailwindcss does, and what are the options?
- Can it be compiled from sources?
- Can we use depext (though I don't see a package on FreeBSD)?
- Does the binary (and/or its configuration) change regularly, or could we check in the generated output?

TheLortex:
tailwindcss is the css library that is used for the website design (https://tailwindcss.com/). The role of the binary is to look at which CSS properties are used, and generate a minimized CSS file that only contains these properties. 

Indeed that's a problem for FreeBSD, there only are builds for Linux, MacOS and Windows. The CLI is written in Js, so it can be executed using NodeJS. I suspect the binaries embed a NodeJS runtime. 

I don't think the binary changes a lot, but the generated output depends on the HTML files so it needs to be refreshed if these files are edited. Checking it in would enable editing the content of the website, but not the template.

hannesm:
Thanks for your reply. Given that "it doesn't change a lot, would it be possible that we setup a separate GitHub action (or whichever CI people prefer that is easy to setup and maintain) which on every PR & push to main generates the talwindcss output, and compares it with the checked in version?

It could even directly push a new version if necessary. It'd allow to remove some complexity from a normal local build. WDYT?

hannesm:
Due to the year-long discussions about mirage-www, and my involvement in other web services, I was wondering what you think about the following path:

(1) We note that mirage-www is mostly a static website. The content is sometimes updated via blog posts / pull requests, but there's no interactive playing field for clients / web browsers.

(2) The code base is slightly complex, since it achieves multiple things: templating, HTTP server, unikernel, layouting.

(3) The data and code is developed together in this git repository, which means a change of the data needs recompilation / redeployment.

Now, from a sustainability and complexity perspective, I don't see a reason to have the content management and the web services integrated. From my experience, e.g. with Canopy https://github.com/Engil/Canopy, I discovered that maintenance of such a project is hard, especially if the projects (opam packages) you depend on are moving (sometimes fast).

[...]

Issue #757, opened 2022-05-11
Build error while running `make build`


davisuga:
OS: Linux Mint 20.02 Uma
OCaml: 4.14
Opam: 2.1.0
Dune: 3.0.3

Output: 

```bash
opam exec -- dune build --root .
File "mirage/duniverse/ocaml-graphql-server/dune-project", line 2, characters 14-17:
2 | (using menhir 2.0)
                  ^^^
Warning: Version 2.0 of the menhir extension is not supported until version
1.4 of the dune language.
Supported versions of this extension in version 1.1 of the dune language:
- 1.0
File "mirage/dune.build", line 14, characters 42-50:
14 |    mirage-runtime mirage-time mirage-unix mirageio tcpip.stack-socket
                                               ^^^^^^^^
Error: Library "mirageio" not found.
-> required by _build/default/mirage/main.exe
-> required by alias mirage/all
-> required by alias default
File "mirage/duniverse/dream/src/vendor/paf/lib/paf.ml", line 345, characters 16-17:
345 |         | Error e ->  Logs.info (fun m -> m "Error: unknown"); Lwt.return ()));
                      ^
Error (warning 27 [unused-var-strict]): unused variable e.
make: *** [Makefile.user:22: build] Error 1
```

d-s-d:
I have the same issue on OS x.

`➜  mirage-www git:(master) ✗ uname -a
Darwin ... 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:23 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T6000 arm64`

`
opam exec -- dune build --root . @install
-e \e[2musing overlay repository mirage: [opam-overlays, mirage-overlays] \e[0m
[opam-overlays] no changes from git+https://github.com/dune-universe/opam-overlays.git
[NOTE] Repository opam-overlays has been added to the selections of switch /.../mirage-www only.
       Run `opam repository add opam-overlays --all-switches|--set-default' to use it in all existing switches, or in newly created switches, respectively.

[mirage-overlays] no changes from git+https://github.com/dune-universe/mirage-opam-overlays.git
[NOTE] Repository mirage-overlays has been added to the selections of switch /.../mirage-www only.
       Run `opam repository add mirage-overlays --all-switches|--set-default' to use it in all existing switches, or in newly created switches, respectively.

 ↳ generate lockfile for monorepo dependencies
==> Using 1 locally scanned package as the target.
[...]

Issue #796, opened 2023-02-22
`Error: No rule found for asset/main.css`


RyanGibb:
When building the project I'm getting the error `Error: No rule found for asset/main.css` because the dependancy https://github.com/mirage/mirage-www/blob/5dc832a346948377eb7f8f0947a4e415d8f4ce17/lib/dune#L11 which I think is broken because of https://github.com/mirage/mirage-www/commit/eafef49b64315194aa6ac6940ebf6c2e0079f5c1.

Should `../asset/main.css` be removed here?

TheLortex:
`asset/main.css` is generated by the following rule, declared in https://github.com/mirage/mirage-www/blob/master/dune

```
(subdir
 asset/
 (rule
  (target main.css)
  (deps
   %{workspace_root}/tailwind.config.js
   %{workspace_root}/template/styles.css
   (glob_files %{workspace_root}/template/*.eml))
  (action
   (chdir
    %{workspace_root}
    (run tailwindcss build -c tailwind.config.js -i template/styles.css -o
      %{target})))))
```

So the error is a bit surprising.

RyanGibb:
Interesting. It looks like the tailwindcss command was silently failing then. It looks like building it from source has fixed this https://github.com/mirage/mirage-www/pull/797/commits/434508c6eaa172dfd3603471b6e32bc095fc4421.

PR #795, opened 2023-02-21
Update dream

TheLortex:
I have enabled ocaml-ci on my dream fork and opened a PR with the sufficient changes to make the CI happy: https://github.com/TheLortex/dream/pull/3

This PR changes the dream commit to point to the updated version.

PR #797, opened 2023-02-22
Add support for building with Nix

RyanGibb:
This PR adds support for building the project with [Nix](https://nixos.org/).

It supports building a binary program as well as a unikernel with https://github.com/tweag/opam-nix and https://github.com/RyanGibb/hillingar respectively.

To build a binary, one can use `$ nix build .` or `$ nix build .#scope.mirageio-bin`.

To build a unikernel `$ nix build .#unikernel.<target>`, e.g. `$ nix build .#unikernel.unix`.

One can also create a development environment from these derivations. So instead of a new user having to go through the manual process of creating an opam switch and installing all the dependencies (along with debugging steps that might arise with the pinned packages, vendored dependencies, and external system packages) they can just run `nix develop .#<derivation> -c <command>`, e.g. ``nix develop . -c codium `pwd` `` or `nix develop .#unikernel.unix -c dune build`.

I hope this will enable reproducible builds to help alleviate issues such as:
- https://github.com/mirage/mirage-www/issues/792
- https://github.com/mirage/mirage-www/issues/757

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -6,6 +6,10 @@
 (rule
  (targets asset.ml asset.mli)
  (deps
+  ; Not specifying dependencies on main.css on purpose so we can skip the generation
+  ; of the CSS file if we only want to run the server.
+  ; Uncomment in development
+  ../asset/main.css
   (source_tree ../asset))
  (action
   (with-stdout-to
--- a/lib/dune
+++ b/lib/dune
@@ -6,10 +6,6 @@
 (rule
  (targets asset.ml asset.mli)
  (deps
-  ; Not specifying dependencies on main.css on purpose so we can skip the generation
-  ; of the CSS file if we only want to run the server.
-  ; Uncomment in development
-  ../asset/main.css
   (source_tree ../asset))
  (action
   (with-stdout-to
--- a/flake.lock
+++ b/flake.lock
@@ -79,11 +79,11 @@
         "opam2json": "opam2json"
       },
       "locked": {
-        "lastModified": 1677106266,
-        "narHash": "sha256-nsujN0ZqX6GCfo7p93CCWFajWXMDDpnHP27emlZikTE=",
+        "lastModified": 1677153290,
+        "narHash": "sha256-yVUoJ1V/WMFC5ZtrWCTuma+sPBHkjewIGHpOvn7N6HU=",
         "owner": "RyanGibb",
         "repo": "hillingar",
-        "rev": "fb9c1eabece8c87a864a3d4bcc23cd2c6d24c57d",
+        "rev": "ae21693a50aa8267bc461b3a680cb93239191c0c",
         "type": "github"
       },
       "original": {
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,273 @@
+{
+  "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1627913399,
+        "narHash": "sha256-hY8g6H2KFL8ownSiFeMOjwPC8P0ueXpCVEbxgda3pko=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "12c64ca55c1014cdc1b16ed5a804aa8576601ff2",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1676283394,
+        "narHash": "sha256-XX2f9c3iySLCw54rJ/CZs+ZK6IQy7GXNY4nSOyu2QG4=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "3db36a8b464d0c4532ba1c7dda728f4576d6d073",
+        "type": "github"
+      },
+      "original": {
+        "id": "flake-utils",
+        "type": "indirect"
+      }
+    },
+    "flake-utils_2": {
+      "locked": {
+        "lastModified": 1676283394,
+        "narHash": "sha256-XX2f9c3iySLCw54rJ/CZs+ZK6IQy7GXNY4nSOyu2QG4=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "3db36a8b464d0c4532ba1c7dda728f4576d6d073",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "flake-utils_3": {
+      "locked": {
+        "lastModified": 1638122382,
+        "narHash": "sha256-sQzZzAbvKEqN9s0bzWuYmRaA03v40gaJ4+iL1LXjaeI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "74f7e4319258e287b0f9cb95426c9853b282730b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "hillingar": {
+      "inputs": {
+        "flake-utils": "flake-utils_2",
+        "nix-filter": "nix-filter",
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "opam-nix": [
+          "opam-nix"
+        ],
+        "opam-overlays": [
+          "opam-overlays"
+        ],
+        "opam-repository": [
+          "opam-repository"
+        ],
+        "opam2json": "opam2json"
+      },
+      "locked": {
+        "lastModified": 1677106266,
+        "narHash": "sha256-nsujN0ZqX6GCfo7p93CCWFajWXMDDpnHP27emlZikTE=",
+        "owner": "RyanGibb",
+        "repo": "hillingar",
+        "rev": "fb9c1eabece8c87a864a3d4bcc23cd2c6d24c57d",
+        "type": "github"
+      },
+      "original": {
+        "owner": "RyanGibb",
+        "repo": "hillingar",
+        "type": "github"
+      }
+    },
+    "mirage-opam-overlays": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1661959605,
+        "narHash": "sha256-CPTuhYML3F4J58flfp3ZbMNhkRkVFKmBEYBZY5tnQwA=",
+        "owner": "dune-universe",
+        "repo": "mirage-opam-overlays",
+        "rev": "05f1c1823d891ce4d8adab91f5db3ac51d86dc0b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "dune-universe",
+        "repo": "mirage-opam-overlays",
+        "type": "github"
+      }
+    },
+    "nix-filter": {
+      "locked": {
+        "lastModified": 1659352118,
+        "narHash": "sha256-X/Tdlj/PYxcQg/1hcHXxdnDr5zLO22LohIudX+oT968=",
+        "owner": "numtide",
+        "repo": "nix-filter",
+        "rev": "3e1fff9ec0112fe5ec61ea7cc6d37c1720d865f8",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "ref": "3e1fff9",
+        "repo": "nix-filter",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1677092149,
+        "narHash": "sha256-FWDw98KGktGY3cOGs0xB7/M28mFrQ6mScHLPSjF0PLs=",
+        "owner": "nixos",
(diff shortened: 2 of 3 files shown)


RyanGibb:
I think the CI is failing because dream is using the wrong version of `mirage-crypto-rng` after `mirage-crypto-rng.lwt` was split into `mirage-crypto-rng-lwt`: https://github.com/mirage/mirage-crypto/commit/0b732b72b607bf4ea593a60fc7327094aaae61a2

```
#=== ERROR while compiling dream.dev ==========================================#
# context     2.1.4 | linux/x86_64 | ocaml-base-compiler.5.0.0 | pinned(git+https://github.com/TheLortex/dream.git#e8ba8e4c6c8689e0e568153fd908369fb8303d53#e8ba8e4c)
# path        ~/.opam/5.0/.opam-switch/build/dream.dev
# command     ~/.opam/5.0/bin/dune build -p dream -j 71
# exit-code   1
# env-file    ~/.opam/log/dream-1-1d9bf4.env
# output-file ~/.opam/log/dream-1-1d9bf4.out
### output ###
# File "src/dune", line 19, characters 2-23:
# 19 |   mirage-crypto-rng.lwt
#        ^^^^^^^^^^^^^^^^^^^^^
# Error: Library "mirage-crypto-rng.lwt" not found.
# -> required by library "dream" in _build/default/src
# -> required by _build/default/META.dream
# -> required by _build/install/default/lib/dream/META
# -> required by _build/default/dream.install
# -> required by alias install
```

Pinning `mirage-crypto-rng = "0.10.7";` fixes this, but I think updating the pinned for of dream is the way to go because https://github.com/TheLortex/dream/commit/fe59e41d8c963e950f81ecd0e3387f6c799d6885 fixed this.

lib/dune
```diff
@@ -6,10 +6,6 @@
 (rule
  (targets asset.ml asset.mli)
  (deps
-  ; Not specifying dependencies on main.css on purpose so we can skip the generation
-  ; of the CSS file if we only want to run the server.
-  ; Uncomment in development
-  ../asset/main.css
```
RyanGibb, line 12: This was removed to allow the build to succeed. See https://github.com/mirage/mirage-www/issues/796.

hannesm:
See #795 for the mirage-crypto-rng 0.11.0 change (and requirements to get this through, i.e. a release of mirage).

I'm sceptical of adding 360 lines to this project (adding to the complexity of it), even more since this code is not covered by CI, and there's no obvious path how to upgrade the lock file (esp. without "nix" tooling - I would be sad if "contributing to mirage-www means you have to have nix tooling and some expertise in that).

This won't fix #792 (for me), since the underlying issue that tailwindcss is distributed as binary for some platforms (not including FreeBSD).

So, thanks for your contribution. I'm happy to enable people to build MirageOS unikernels with NixOS, but I don't think spreading lock files around is a viable path (e.g. we have this in albatross (https://github.com/roburio/albatross) - even with a CI job that attempts to build it - but there's lack of automation on updating the lock file).

I personally like to build (and see failures and fix them) with the latest HEAD of opam-repository, instead of pointing to specific commits and find out weeks, months, years later that there's a huge amount of technical debt accumulated. But that's of course a debatable development methodology.

RyanGibb:
> I'm sceptical of adding 360 lines to this project (adding to the complexity of it) ... I would be sad if "contributing to mirage-www means you have to have nix tooling and some expertise in that).

That is a valid concern. The nice thing about opam-nix is that it continues to work with the existing opam tooling, so it allows those using opam to collaborate with those using nix.

> This won't fix #792 (for me), since the underlying issue that tailwindcss is distributed as binary for some platforms (not including FreeBSD).

I believe https://github.com/mirage/mirage-www/pull/797/commits/434508c6eaa172dfd3603471b6e32bc095fc4421 should remedy this. It builds tailwindcss from source, and Nix is [available](https://wiki.freebsd.org/Nix) for FreeBSD (although I've never used it myself). Nix really shines in handling these kind of multi-language environments.

> So, thanks for your contribution. I'm happy to enable people to build MirageOS unikernels with NixOS, but I don't think spreading lock files around is a viable path (e.g. we have this in albatross (https://github.com/roburio/albatross) - even with a CI job that attempts to build it - but there's lack of automation on updating the lock file).

Regarding the lack of support for automatic upgrading of lockfiles, this is perhaps an intentional feature of Nix providing reproducibility. I imagine it would be possible to set up CI to automatically upgrade this, but any breaking changes would need to be dealt with manually (but this is also the case outside of Nix).

> I personally like to build (and see failures and fix them) with the latest HEAD of opam-repository, instead of pointing to specific commits and find out weeks, months, years later that there's a huge amount of technical debt accumulated. But that's of course a debatable development methodology.

One can in fact do this with opam-nix too. Removing `ocaml-base-compiler = "*"` from the query should get you the trunk which you can evaluate with `--impure` due to the lack of a static hash. I've submitted this PR without the latest head in order to enable reproducibility, however.

hannesm:
Thanks for your reply @RyanGibb, still I've a couple of questions unanswered:
- why should I buy into nix and try it out on FreeBSD? I'm already at the complexity budget using opam, mirage, opam-monorepo -- now on top I should use nix? if it would remove some of the complexity of the other moving parts, that would be nice, but as far as I understand it just adds another layer
- the file you would like to add is not tested by any CI system (neither the deployment CI, nor the ocaml-ci use nix)
- it is still unclear how this lock file would be updated

I understand that nix uses lock files for reproducible builds and binaries, but as I tried to emphasize in the OCaml & opam ecosystem there are rolling releases of sources, and I find it crucial to stay on top of the releases. I've worked on a thin extension for opam to feature reproducible builds (https://github.com/roburio/orb -- https://builds.robur.coop), which uses the HEAD of opam-repository (and indeed there are lots of updates to opam-repository that don't result in different hashes of the resulting unikernel binaries) -- so there's no requirement to update every now and then commit hashes -- but instead failures can be seen early (and then fixed).

So, I wonder how to join forces or document/explain more clearly what the development model of the mirage-www unikernel (and other unikernels) is -- and from what I read there is a divergence between "let's stick to a certain opam-repository commit" and "let's try and fail the latest, observe CI and fix the code".

On a more technical level, in your lock file, there is twice `opam-overlays` (and `opam-overlays2`) with different `rev`, also twice `opam2json` (and `opam2json2`) -- this time with the same `rev` -- are you aware of that? Is that intentional?

hannesm:
And to be clear, I'm open to have this merged if others in the @mirage/core team think this is a good path to have. I'm curious what others think in terms of development model for MirageOS unikernels. I'm just hesitant since several years ago it was "let's add a debian subdirectory with changelog, dependencies, ...", then some years ago it was "let's add Dockerfile into each repository", and now it is lock files for various systems (nix, npm, esy, ...).

dinosaure:
I would like to extend the @hannesm's point of view by the current state of `mirage-www`. The project is currently not in a good shape when it requires a certain fork of some libraries that we don't have the time to maintain (this is not a criticism here, as I also fit in with those who do not have the time to maintain such forks). This means above all that, as it stands, `mirage-www` already has serious problems in maintaining itself and in the time available to everyone to keep the project usable.

Adding support for NixOS implies not only a responsibility on your part to maintain that support but also a collective responsibility to those involved in the MirageOS project to keep an eye on that support.

This is, I think, where things become problematic and can be misinterpreted. Like @hannesm, I am not against NixOS support and we have repeatedly shown good signals to the NixOS community to improve MirageOS libraries and software (like Solo5). However, we are not yet in a position to systematically (and for obvious reasons of time) integrate NixOS issues into all our projects.

What would be interesting in this context is to be able to support NixOS semi-automatically from a well-defined tool or workflow - note that MirageOS has already experimented with this kind of transition in the past (notably in the support of AFL). If such a thing exists and can be easily appropriated collectively, I would personally be happy to experiment/test and go further in this direction.

RyanGibb:
Hi @hannesm and @dinosaure. First of all, thanks for taking the time and the energy to respond.

To respond to @hannesm first:

> still I've a couple of questions unanswered:
> - why should I buy into nix and try it out on FreeBSD? I'm already at the complexity budget using opam, mirage, opam-monorepo -- now on top I should use nix? if it would remove some of the complexity of the other moving parts, that would be nice, but as far as I understand it just adds another layer

I should preface this by saying that while I find it to be very useful, I'm not trying to convince you yourself to use Nix if you don't have the time or energy to learn this new tool.

I suppose the simplest answer would be that it addresses your issue of building tailwindcss from source. I would also argue that instead of adding another layer on top of all these tools it instead replaces a lot of functionality of Opam with a philosophy I find easier to work with. Rather than creating and maintaining mutable Opam switches, you can instead use Nix to deterministically build and provide sources for development and builds. You can transparent share these dependencies between projects, garbage collect them, and recreate them trivially. I would say it remove a lot of the complexity of Opam; the only part of Opam that is used for in `opam-nix` is solving the dependency constraints that `.opam` files provide.

While I find it to be very useful, people like tools they are used to, and it is unarguably additional work to gain familiarity with Nix. I would argue not because it's necessarily more complex but rather because it's unfamiliar.

> - the file you would like to add is not tested by any CI system (neither the deployment CI, nor the ocaml-ci use nix)

I don't have any experience with `mirage-www`'s CI, but I could look into setting something up. This would perhaps also help address:

> - it is still unclear how this lock file would be updated
> 
> I understand that nix uses lock files for reproducible builds and binaries, but as I tried to emphasize in the OCaml & opam ecosystem there are rolling releases of sources, and I find it crucial to stay on top of the releases. I've worked on a thin extension for opam to feature reproducible builds (https://github.com/roburio/orb -- https://builds.robur.coop/), which uses the HEAD of opam-repository (and indeed there are lots of updates to opam-repository that don't result in different hashes of the resulting unikernel binaries) -- so there's no requirement to update every now and then commit hashes -- but instead failures can be seen early (and then fixed).

It seems like a CI action or GitHub action that runs `nix flake update` (or `nix flake lock --update-input opam-repository`) could provide the benefits of both. The benefit of having such a lockfile in your project is that you essentially track the external dependencies to the exact hashes along with the source code. You could in fact not commit the lockfile and recreate it on every build, but this might lead to incompatibilities if breaking changes were made to one of the flake inputs (e.g. opam-nix).

> So, I wonder how to join forces or document/explain more clearly what the development model of the mirage-www unikernel (and other unikernels) is -- and from what I read there is a divergence between "let's stick to a certain opam-repository commit" and "let's try and fail the latest, observe CI and fix the code".

I am afraid I'm unfamiliar with how other unikernels are developed in general. I can certainly see the advantage to the former though.

> On a more technical level, in your lock file, there is twice opam-overlays (and opam-overlays2) with different rev, also twice opam2json (and opam2json2) -- this time with the same rev -- are you aware of that? Is that intentional?

This is the result of different projects lockfiles composing. By default it's conservative and duplicates them, but it's possible to deduplicate them. I'll push an update doing this now.

---

And to respond to @dinosaure:

> I would like to extend the @hannesm's point of view by the current state of mirage-www. The project is currently not in a good shape when it requires a certain fork of some libraries that we don't have the time to maintain (this is not a criticism here, as I also fit in with those who do not have the time to maintain such forks). This means above all that, as it stands, mirage-www already has serious problems in maintaining itself and in the time available to everyone to keep the project usable.

I understand your problem, and empathise with it.

> Adding support for NixOS implies not only a responsibility on your part to maintain that support but also a collective responsibility to those involved in the MirageOS project to keep an eye on that support.
> 
> This is, I think, where things become problematic and can be misinterpreted. Like @hannesm, I am not against NixOS support and we have repeatedly shown good signals to the NixOS community to improve MirageOS libraries and software (like Solo5). However, we are not yet in a position to systematically (and for obvious reasons of time) integrate NixOS issues into all our projects.

My motivation in this PR is not to add support for NixOS (or else I would have also added a NixOS module for deployment on a NixOS system), but rather simply support to build the project with Nix for the benefits in reproducible builds and development environments. Nix, for our purposes here, is a build system and package manager. And NixOS is a Linux distribution built on top of Nix. I've written a bit about the differences between [Nix](https://ryan.freumh.org/blog/hillingar/#nix) and [NixOS](https://ryan.freumh.org/blog/hillingar/#nixos). However, I would be happy to help to maintain this Nix support.

I hope this clarifies that I do not aim to support an additional operating system with this, but rather integrate a useful tool.

> What would be interesting in this context is to be able to support NixOS semi-automatically from a well-defined tool or workflow - note that MirageOS has already experimented with this kind of transition in the past (notably in the support of AFL). If such a thing exists and can be easily appropriated collectively, I would personally be happy to experiment/test and go further in this direction.

Regarding semi-automatic support, I'm not entirely sure how this would work. I think the closest existing project is https://github.com/RyanGibb/hillingar which aims to be an interface to simplify adding support for building Mirage unikernels with Nix. This `flake.nix` is really just a simply wrapper calling functionality in hillingar for building a unikernel and opam-nix for building a package and providing a development environment, alongside some fixes to work around quirks of the project. 

In summary, I think the benefits of this PR is support for reproducible development environments and deployments.

RyanGibb:
Reading https://github.com/leanprover/lake/pull/15#issuecomment-927186245 made me realise that an alternative might be just to omit committing the lockfile.

RyanGibb:
However I can see there isn't much interest in this so I'm going to close the issue

PR #804, opened 2023-05-03
Fix `make switch`

pkel:
I propose three changes:
1. Update `dream` version to a recent commit at https://github.com/aantron/dream. This fixes a missing dependency on `rresult`.
2. Simplify implementations of `make switch` and `make deps` to be use command only. The problem with the previous list of commands was that `opam install ocaml-lsp-server` resulted in an upgrade of ocaml from version 4 to 5 and recompilation of the whole switch.
3. `make fmt` after `make deps` was complaining about a wrong version of `ocamlformat`. I now fix this version in `dune-project`.

With these changes I can build and run the mirage website from a clean checkout with `make switch start`.


dune-project
```diff
@@ -45,3 +45,13 @@
...
+
+(package
+ (name _tools)
+ (synopsis "Dev tools for mirage-www")
+ (description "Dev tools for mirage-www")
+ (depends
+  dune-release
+  (ocamlformat (= 0.24.1))
+  utop
+  ocaml-lsp-server))
```
kit-ty-kate, line 57:
```suggestion
  (ppx_deriving_yaml :build))
  (dune-release (and :with-test :dev))
  ((ocamlformat (and (= 0.24.1) :with-test :dev))
  (utop (and :with-test :dev))
  (ocaml-lsp-server (and :with-test :dev))))
```

tmcgilchrist:
Deployability check fix and minor upgrade of the Dream versions on https://github.com/mirage/mirage-www/pull/805. @kit-ty-kate

samoht:
Thanks - I've pushed a commit to simplify the Makefile instead and not install the full list of platform tools (we don't need dune-release here for instance).

PR #805, opened 2023-05-08
Update to Dream.1.0.0~alpha5 plus dream-mirage fixes.

tmcgilchrist:
The opam pins match those for `ocaml-matrix` and the Dockerfile updates are required for the deployability check on [deploy.mirage.io](https://deploy.mirage.io)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM ocaml/opam:debian-11-ocaml-4.13
+FROM ocaml/opam:debian-12-ocaml-4.14
+RUN sudo apt-get update && sudo apt-get install autoconf automake -y --no-install-recommends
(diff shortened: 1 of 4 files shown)


hannesm:
sorry, I don't have any knowledge about it (neither the build system used in this repository nor how upgrading dream should be done).

samoht:
@tmcgilchrist thanks! The CI picked an issue with lower bounds for `mimic`: could you add the missing constraint in the pinned opam file?

TheLortex:
Something changed because `autoreconf` is not in path anymore. `stdcompat` uses it but doesn't declare a dependency on `conf-autoconf`.

samoht:
I've force-pushed a fix.

PR #817, opened 2024-04-07
 Update examples in tutorial-lwt.md with current versions. Fix a couple of issues in text.

oemmerson:
The examples in 'Getting Started with Lwt threads' tutorial on the website are outdated and do not compile on the current Mirage.
I have replaced the code with the current buildable versions from https://github.com/mirage/mirage-skeleton/tree/main/tutorial/lwt and edited the accompanying text where needed.

I also updated a couple of links to documentation in the text to the latest versions to be consistant and added a small clarifier.

### DIFF
--- a/data/dune
+++ b/data/dune
@@ -8,6 +8,7 @@
 
 (library
  (name mirageio_data)
+ (public_name mirageio.data)
  (libraries ptime)
  (modules mirageio_data))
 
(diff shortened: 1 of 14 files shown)


samoht:
That looks great thanks. Let me update this to the latest `mirage` and see if we can automate checking that those examples keep compiling.

samoht:
Ok I've pushed a tentative use of `mdx` to check the lwt tutorial with the CI. If that works I'll try to port more of them :-)

PR #818, opened 2024-04-09
update to mirage 4.5.0

samoht:


mirage/config.ml
```diff
@@ -155,24 +82,35 @@ let internal_stack =
...
 let mirage_monitoring =
   let ip_key =
     let ip =
       let doc =
-        Key.Arg.info ~doc:"IP of InfluxDB server to transmit metrics to"
+        Arg.info ~doc:"IP of InfluxDB server to transmit metrics to"
           [ "metrics-ip" ]
       in
-      Key.(create "metrics-ip" Arg.(required ip_address doc))
+      Key.create "metrics-ip" (required ip_address doc)
```
samoht, line 103: @hannesm: what's the intended use of this key? This is now failing at configuration time if you don't provide a `--metrics-ip`, which I suspect is not what is expected. Should we turn those keys to runtime keys?

hannesm: I guess that is the right path to take. The "intended use" is described above, "IP of InfluxDB server to transmit metrics to" -- so yes this should be a runtime key.

hannesm: https://github.com/ocaml/opam-repository/pull/25649

samoht:
pfff, all green! I'll cleanup and release the right things now

Issue #790, opened 2023-01-04
http://mirage.io/xyz already redirects to https://mirage.io (the path is lost on redirections)


kit-ty-kate:
Steps to reproduce:
1) Go to http://mirage.io/xzy
2) You are now on https://mirage.io instead of the expected https://mirage.io/xzy

TheLortex:
That's because of that piece of code which doesn't look at the request's path:
https://github.com/mirage/mirage-www/blob/5c5a51ecdf8ecf6837fa10efa2b9497596603d24/mirage/unikernel_tls.ml#L54

samoht:
This seems to be fixed. Please re-open if not.

hannesm:
At least for me, this doesn't seem to be fixed. I browsed to http://mirage.io/xyz and ended up at https://mirage.io

samoht:
Same here, that's very odd, both Chrome and Safari gives me a proper 404

hannesm:
```
$ curl -v http://mirage.io/xyz
* Host mirage.io:80 was resolved.
* IPv6: (none)
* IPv4: 147.75.34.212
*   Trying 147.75.34.212:80...
* Connected to mirage.io (147.75.34.212) port 80
> GET /xyz HTTP/1.1
> Host: mirage.io
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Location: https://mirage.io
< Content-Length: 0
< 
* Connection #0 to host mirage.io left intact
```

And, as outlined in the initial comment, the "Location" should be "https://mirage.io/xyz".

hannesm:
fixed in #824

PR #824, opened 2024-04-15
Keep the URI targets in the http -> https redirections

samoht:
Tentative fix for #790

I've just tested locally the `unikernel.ml` one.


samoht:
@sabine do you think that's the right fix?

hannesm:
thanks a lot

PR #825, opened 2024-04-23
Update "Hello World" tutorial text to reflect changes from #823

sabine:
The Hello World tutorial text would explain `Keys`, while the code examples now use `runtime_args`. Here, I make an attempt at improving the text to match the code examples.

PS: Hi from MirageOS retreat!


samoht:
Many thanks for all the nice PRs!

PR #826, opened 2024-04-23
Remove Tracing & Profiling Tutorial

sabine:
According to the changelog, the `tracing` feature has been removed in Mirage 4.0, so the tutorial cannot be updated.


samoht:
Has this been removed completely or just changed?

hannesm:
the profiling has been removed completely. thanks for this PR, imho good to go. there'll likely be a revival once runtime events are around :)

samoht:
It looks good then! I'd be happy to merge all those PRs quickly, but GitHub is super slow to compute if the branches are mergeable (every check takes a few minutes!!).

PR #827, opened 2024-04-23
Link RSS feed in head of every page, add RSS icon in sidebar

### LINKED ISSUE #782: new website: atom / rss feed?
Hi,

I couldn't find any advertisement of the feed.xml on the new website (I had to look into the code to figure out that there is still an atom feed)... Shouldn't there be a `<link rel="alternate" href="/feed.xml" title="MirageOS" type="application/atom+xml"/>` in the `<head>` of the main landing page?

Since I don't know much about web, it would be great if someone would chime in.

sabine:
Resolves #782.

RSS feed is now advertised in `head` of every page to make it easier to discover.

UI Change:
|before|after|
|-|-|
|Screenshot 2024-04-23 at 19-21-12 MirageOS Blog|Screenshot 2024-04-23 at 19-21-23 Welcome to MirageOS|

### DIFF
--- a/asset/main.css
+++ b/asset/main.css
@@ -1489,6 +1489,10 @@ video {
   justify-content: space-between;
 }
 
+.gap-3 {
+  gap: 0.75rem;
+}
+
 .space-x-2 > :not([hidden]) ~ :not([hidden]) {
   --tw-space-x-reverse: 0;
   margin-right: calc(0.5rem * var(--tw-space-x-reverse));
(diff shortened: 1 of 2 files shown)


samoht:
Thanks! You might want to force the update of the main CSS file too. We probably need to add a way to do this easily (maybe a dune target?)

The main CSS file is vendored as the tailwind binary is not available on all distributions

PR #839, opened 2024-07-12
Use variant-based bootvar for mdx

Firobe:
To avoid CI errors like [these](https://ocaml.ci.dev/github/mirage/mirage-www/commit/3589405f2f853b4b32209d968670284008fef815/variant/debian-12-5.2_opam-2.1#L788-788)


hannesm:
clearl correct, ci passes. thanks a lot

PR #840, opened 2024-08-23
Deploy with OCaml 5.2 on next

Firobe:
This is exactly https://github.com/mirage/mirage-www/pull/838 containing the work of @shym but tailored to be deployed on branch `next`. Its general purpose is to deploy a version of `mirage-www` built with ocaml-solo5 for 5.2 to benchmark its resource usage.

This PR:
- drops the Github Action CI config from #838
- and instead modifies the `Dockerfile` used by the mirage deployer:
  - bumps the OCaml version
  - pins @shym's version of `ocaml-solo5`
  - hardcodes the `EXTRA_FLAGS` passed by the mirage deployer to remove `--metrics`, since `mirage-memtrace` isn't readily available for OCaml 5.2

That last modification is clearly a hack and should be addressed by either somehow porting `mirage-memtrace` to 5.2 or conditionally passing the right EXTRA_FLAGS in the mirage deployer, but this should be enough to do some benchmarks on the `next` branch.

This otherwise syncs `next` with `master` (dropping commits from https://github.com/mirage/mirage-www/pull/791/commits which have remained there in limbo for quite a while). Hence this will need to be force-pushed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,20 @@
-FROM ocaml/opam:debian-12-ocaml-4.14
+FROM ocaml/opam:debian-12-ocaml-5.2
 RUN sudo apt-get update && sudo apt-get install autoconf automake -y --no-install-recommends
 RUN mkdir -p /home/opam/www/mirage
 WORKDIR /home/opam/www
 RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam
-RUN cd ~/opam-repository && git pull origin master && git reset --hard 6e53e4603d4bb696e4fa1cdbde61b156a3f8e01e
+RUN cd ~/opam-repository && git pull origin master && git reset --hard 705cfbfe709bba2533a3cd6164b7dab37fca0342
 RUN opam update
 RUN opam install 'mirage>=4.5.0'
 COPY --chown=opam:root mirage/config.ml /home/opam/www/mirage/
 COPY --chown=opam:root mirageio.opam /home/opam/www/
 ARG TARGET=unix
 ARG EXTRA_FLAGS=
-RUN opam exec -- mirage configure -f mirage/config.ml -t $TARGET $EXTRA_FLAGS
+ARG EXTRA_FLAGS_NO_METRICS="--tls=true --separate-networks"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,18 +1,20 @@
...
+RUN sudo ln -f /usr/bin/opam-2.1 /usr/bin/opam
+RUN cd ~/opam-repository && git pull origin master && git reset --hard 705cfbfe709bba2533a3cd6164b7dab37fca0342
+RUN opam update
+RUN opam install 'mirage>=4.5.0'
 COPY --chown=opam:root mirage/config.ml /home/opam/www/mirage/
 COPY --chown=opam:root mirageio.opam /home/opam/www/
 ARG TARGET=unix
 ARG EXTRA_FLAGS=
-RUN opam exec -- mirage configure -f mirage/config.ml -t $TARGET $EXTRA_FLAGS
+ARG EXTRA_FLAGS_NO_METRICS="--tls=true --separate-networks"
```
shym, line 13: I wondered why not use `EXTRA_FLAGS` instead of creating the `EXTRA_FLAGS_NO_METRICS`.
Or, taken the other way around, if keeping `EXTRA_FLAGS` separate is useful, shouldn’t it still appear in the configuration command lines?

shym (approved):
Apart from a small comment I wrote on the corresponding line, this looks good to me!

Dockerfile
Firobe: Basically, the CI system always passes `--metrics` through `EXTRA_FLAGS` when building `mirage-www` (and I don't have control over that CI system). And I couldn't find a reasonable way to do something like `EXTRA_FLAGS_NO_METRICS = $EXTRA_FLAGS without --metrics`.

Another way would be for this branch to still accept --metrics but disable the underlying behavior (and not pull the dependency).

I went with the hacky solution (to hardcode what the CI passes without --metrics) to modify the code as little as possible so this experiment is representative. Then the correct solution will be to change the CI system to pass the correct flags, once we know we want to do this on `main`.

samoht:
I've force-pushed this to the `next` branch. Let's see if the world explodes...

PR #851, opened 2025-04-22
Resurrect /security

### LINKED ISSUE #778: the security advisories contain a link to mirage.io/securty which returns a 404
Is there a way with this shiny new website to have the content for /security back as it was? it is at the moment available under https://mirage.io/docs/security, but since the security advisories point to /security, it would be great to keep that url working as well.

samoht:
Really fix #778 this time.

### DIFF
--- a/lib/mirageio.ml
+++ b/lib/mirageio.ml
@@ -166,7 +166,7 @@ module Make (Stack : Tcpip.Stack.V4V6) = struct
         Dream.get "/blog/:permalink" Handler.blog_inner;
         Dream.get "/community" Handler.community;
         (* It's important to keep /security at the root as it's linked in
-           securty advisories that are signed (and so can't be changed). *)
+           security advisories that are signed (and so can't be changed). *)
         Dream.get "/security" Handler.security;
         Dream.get "/docs/security" Handler.security_redirect;
         Dream.get "/docs" Handler.docs;


hannesm:
The main goal looks fine. Just for understanding: we put the content under /security again and then have /docs/security being a redirect.

would it make sense to add some comment(s)? do you plan to make the CI green?

samoht:
I'm trying to fix the CI but I get some weirds errors when updating the wiki to mirage 4.9:

```
```bash dir=files/mirage-skeleton/tutorial/noop
$ mirage configure -t unix
Successfully configured the unikernel. Now run 'make' (or more fine-grained steps: 'make all', 'make depends', or 'make lock').
$ dune build
File "lib/devices/random.ml", line 13, characters 2-37:
Error: Unbound value Mirage_crypto_rng_mirage.initialize
[1]
```

hannesm:
Now that @Willenbrink updated the MirageOS website, you may want to rebase on the current master branch and see if the CI passes.

A remaining question is which path should return a redirect, and which redirect (temporary/permanent)?

samoht:
I've rebased

I think keeping `/security` as the main path "looks" better (and so `docs/security` should be a permanent redirect like this patch do). But really no strong opinion.

hannesm:
looks fine to me (apart from the minor typo), please merge

